PR #2, opened 2019-07-14
Add handling events by navigators not only on the top of the tree

osdnk:


src/NavigationContainer.tsx
```diff
@@ -13,6 +13,7 @@ const MISSING_CONTEXT_ERROR =
 
 export const NavigationStateContext = React.createContext<{
   state?: NavigationState | InitialState;
+  getKey: () => string | null;
```
satya164: Instead of `getKey`, it maybe better to just pass `key` as a property

satya164: Maybe use `undefined` instead of `null`? xd

satya164 (commented):
It'll be good to extract out the child listener stuff to a separate hook

src/NavigationContainer.tsx
osdnk: I'd rather stay with `getKey` for a while. Idk why I cannot change it.

osdnk:
Could you take a look on tests pls? 😄 Sth's wrong

satya164:
I have rebased the PR against master.

Regarding tests, looks like a legit issue, no? The child state doesn't seem to be initialised properly in the global state.

I found another issue:

Kapture 2019-07-16 at 13 01 03

There's probably some stale state being updated somewhere. I think moving state to local will make it easier to handle.

src/useDescriptors.tsx
```diff
@@ -52,6 +66,7 @@ export default function useDescriptors({
                 screen={screen}
                 getState={getState}
                 setState={setState}
+                getKey={() => route.key}
```
satya164: If we pass it inline here, it'll invalidate the memo for `navigation` prop in `SceneView` every render.

src/useChildrenActionHelper.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const index = childrenOnActionListeners.indexOf(listener);
+    childrenOnActionListeners.splice(index, 1);
+  };
+
+  React.useEffect(() => {
+    parentRegisterListener && parentRegisterListener(onAction);
+    return () => {
+      parentUnregisterListener && parentUnregisterListener(onAction);
+    };
+  }, [onAction, parentRegisterListener, parentUnregisterListener]);
```
satya164: Need to memo `registerListener` and `registerListener`. Otherwise this effect will run every render in children as these functions are re-created every render.

src/useDescriptors.tsx
satya164: Though if all these function does is `() => route.key`, shouldn't using `route.key` directly work? 🤔

src/useNavigationBuilder.tsx
```diff
@@ -92,10 +99,22 @@ export default function useNavigationBuilder(
     [getCurrentState, router.getRehydratedState, router.getInitialState]
   );
 
+  const { childrenOnActionListeners } = React.useMemo<{
```
satya164: We need ti use `React.useRef` for mutable values like this. `useMemo` isn't guaranteed to persist every render.

src/useOnAction.tsx
```diff
@@ -35,13 +59,38 @@ export default function useOnAction({
...
       if (handleActionParent !== undefined) {
         // Bubble action to the parent if the current navigator didn't handle it
-        if (handleActionParent(action)) {
+        if (handleActionParent(action, state.key)) {
           return true;
         }
       }
 
+      if (
+        router.shouldActionPropagateToChildren &&
```
satya164: I think this method should always exist as part of router's signature

src/useOnAction.tsx
```diff
@@ -1,32 +1,56 @@
...
+
+      if (sourceNavigatorKey === state.key) {
+        return false;
+      }
       const result = getStateForAction(state, action);
 
       if (result !== null) {
-        if (state !== result) {
+        if (handleChildUpdateParent) {
+          const shouldFocus = router.shouldActionChangeFocus
```
satya164: I think this method should always exist as part of router's signature

satya164:
I have fixed the tests and refactored the code a bit. Please take a look when you get time 😁

satya164 (approved):

satya164:
Merged in 5eb12aad812e92278cfcffb59a219fc9a56a2821

PR #3, opened 2019-07-15
Add createNavigator and not export Screen directly

osdnk:


example/index.tsx
```diff
@@ -22,16 +20,6 @@ type TabParamList = {
   fifth: undefined;
 };
 
-const Stack: TypedNavigator<StackParamList, typeof StackNavigator> = {
```
satya164, line 25: I think we still want annotation for our types to work:

```ts
const MyStack: TypedNavigator<StackParamList, typeof StackNavigator> = Stack;
```

example/TabNavigator.tsx
```diff
@@ -143,7 +144,7 @@ const TabRouter: Router<Action | CommonAction> = {
   },
 };
 
-export default function TabNavigator(props: Props) {
+export default createNavigator((props: Props) => {
```
satya164: I think better to do `export default createNavigator(TabNavigator)` below so this component has a display name. Otherwise it'll show up as `Unnamed` in dev tools and `anonymous` in stack traces.

example/StackNavigator.tsx
```diff
@@ -7,7 +7,7 @@ import {
   NavigationProp,
   CommonAction,
   ParamListBase,
-  Router,
+  Router, createNavigator,
```
satya164: Shouldn't prettier throw here 🤔

src/createNavigator.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import { ParamListBase, ScreenProps, TypedNavigator } from './types';
+import Screen from './Screen';
+
+export default function createNavigator<ParamList extends ParamListBase>(
+  Navigator: React.ComponentType<any>
```
satya164: I think we need to use generic for this for types to work properly

```
<ParamList extends ParamListBase, N extends React.ComponentType<any>>(Navigator: N): TypedNavigator<ParamList, N>
```

Though since this function is called by navigator now, we need to change it to not include the param list stuff.

satya164 (changes requested):

example/index.tsx
osdnk: yeah, but they doesn't stop working?

satya164 (approved):

PR #6, opened 2019-07-17
feat: move navigation.state to a route prop

osdnk:


src/SceneView.tsx
```diff
@@ -35,9 +35,8 @@ export default function SceneView(props: Props) {
           ),
         });
       },
-      state: route,
     }),
-    [getState, helpers, route, setState]
+    [getState, helpers, setState]
```
satya164: Doesn't `route.key` need to be in deps?

satya164 (changes requested):

example/index.tsx
```diff
@@ -26,14 +26,16 @@ const MyTab = Tab<TabParamList>();
 
 const First = ({
   navigation,
+  route,
 }: {
   navigation: CompositeNavigationProp<
     StackNavigationProp<StackParamList, 'first'>,
```
osdnk, line 33: fix this typings

PR #16, opened 2019-07-20
feat: make NAVIGATE and JUMP_TO to support key and name of the route

osdnk:


example/StackNavigator.tsx
```diff
@@ -151,25 +151,38 @@ const StackRouter: Router<CommonAction | Action> = {
         });
 
       case 'NAVIGATE':
-        if (state.routeNames.includes(action.payload.name)) {
+        if (
+          action.payload.key ||
```
satya164, line 155: Maybe we should also check if the `key` exists in the stack

src/__tests__/index.test.tsx
```diff
@@ -60,6 +60,76 @@ it('initializes state for a navigator on navigation', () => {
   });
 });
 
+it('throws if NAVIGATE dispatched with both key and name', () => {
```
satya164: Let's move these to a separate file

src/BaseActions.tsx
```diff
@@ -19,8 +19,20 @@ export function goBack(): Action {
...
 }
 
-export function navigate(name: string, params?: object): Action {
-  return { type: 'NAVIGATE', payload: { name, params } };
+export function navigate(target: TargetForAction, params?: object): Action {
+  if (
+    (target.hasOwnProperty('key') && target.hasOwnProperty('name')) ||
+    (!target.hasOwnProperty('key') && !target.hasOwnProperty('name'))
+  ) {
+    throw new Error(
```
satya164, line 27: Maybe instead of throwing in the action creator, we can throw in the router. Then it'll also prevent someone from calling `dispatch` directly with incorrect stuff

src/BaseActions.tsx
```diff
@@ -19,8 +19,20 @@ export function goBack(): Action {
   return { type: 'GO_BACK' };
 }
 
-export function navigate(name: string, params?: object): Action {
-  return { type: 'NAVIGATE', payload: { name, params } };
+export function navigate(target: TargetForAction, params?: object): Action {
```
satya164: We also need to change the types in the `NavigationProp` object in `types.tsx`

example/TabNavigator.tsx
```diff
@@ -171,8 +181,20 @@ const TabRouter: Router<Action | CommonAction> = {
   },
 
   actionCreators: {
-    jumpTo(name: string, params?: object) {
-      return { type: 'JUMP_TO', payload: { name, params } };
+    jumpTo(target: TargetForAction, params?: object) {
```
satya164: We also need to change types for the `TabNavigationProp` type above

satya164 (changes requested):

src/types.tsx
```diff
@@ -2,6 +2,8 @@ import * as BaseActions from './BaseActions';
 
 export type CommonAction = BaseActions.Action;
 
+export type TargetForAction = string | { name: string } | { key: string };
```
satya164: Maybe call it `TargetRoute`?

src/BaseActions.tsx
osdnk: I don't think so. We don't want anyone to dispatch `NAVIGATE` directly. If yes, risk included. I don't want to copy this error to every new and existing navigators.

example/StackNavigator.tsx
osdnk: we do it line 180

satya164 (approved):

PR #18, opened 2019-07-20
feat: add targer argument to setParams

osdnk:


src/BaseActions.tsx
```diff
@@ -13,24 +13,28 @@ export type Action =
...
-    (target.hasOwnProperty('key') && target.hasOwnProperty('name')) ||
-    (!target.hasOwnProperty('key') && !target.hasOwnProperty('name'))
-  ) {
-    throw new Error(
-      'While calling navigate you need to specify either name or key'
-    );
-  }
   if (typeof target === 'string') {
     return { type: 'NAVIGATE', payload: { name: target, params } };
   } else {
```
osdnk, line 29: It was a mistake before. Sorry!

src/BaseActions.tsx
```diff
@@ -42,3 +46,20 @@ export function replace(name: string, params?: object): Action {
 export function reset(state: PartialState & { key?: string }): Action {
   return { type: 'RESET', payload: state };
 }
+
+export function setParams(
+  params: object,
+  target: { name?: string; key?: string }
```
satya164, line 52: I think we should make the signature same as `navigate` etc. I think it's fine if it's required.

src/BaseRouter.tsx
```diff
@@ -0,0 +1,50 @@
+import { CommonAction, Router } from './types';
+
+const BaseRouter: Omit<
+  Omit<Omit<Router<CommonAction>, 'getInitialState'>, 'getRehydratedState'>,
+  'getRehydratedState'
```
satya164, line 5: omitting twice

I think we can remove the annotation. it'll be simpler. If we annotate the individual methods, the type of the object should be inferred.

src/BaseRouter.tsx
```diff
@@ -0,0 +1,50 @@
...
+            r.key === action.payload.key || r.name === action.payload.name
+              ? { ...r, params: { ...r.params, ...action.payload.params } }
+              : r
+          ),
+        };
+      default:
+        return null;
+    }
+  },
+  getStateForRouteNamesChange(state, { routeNames }) {
```
satya164, line 23: nit: I like new lines between methods :)

src/BaseRouter.tsx
```diff
@@ -0,0 +1,50 @@
...
+  },
+  shouldActionPropagateToChildren(action) {
+    return action.type === 'NAVIGATE';
+  },
+
+  shouldActionChangeFocus(action) {
+    return action.type === 'NAVIGATE';
+  },
+
+  actionCreators: {},
```
satya164: since this isn't used, maybe we don't need it?

example/StackNavigator.tsx
```diff
@@ -53,7 +53,6 @@ export type StackNavigationProp<
    */
   popToTop(): void;
 };
-
```
satya164: nit: keep newline

example/StackNavigator.tsx
```diff
@@ -53,7 +53,6 @@ export type StackNavigationProp<
...
 const StackRouter: Router<CommonAction | Action> = {
```
satya164, line 57: The router should also merge base router no? Then we can remove the methods those are the same

src/BaseRouter.tsx
```diff
@@ -0,0 +1,50 @@
...
+      ...state,
+      routeNames,
+      routes: state.routes.filter(route => routeNames.includes(route.name)),
+    };
+  },
+
+  getStateForRouteFocus(state, key) {
+    const index = state.routes.findIndex(r => r.key === key);
+
+    if (index === -1) {
```
satya164:
```suggestion
    if (index === -1 || index === state.index) {
```

osdnk: why?

src/BaseRouter.tsx
osdnk: but it's really simple, isn't it?

src/BaseRouter.tsx
osdnk: I made it optional

src/BaseRouter.tsx
satya164: we don't need to update state if that route was already focused

satya164 (approved):

PR #19, opened 2019-07-21
feat: add a setOptions method to set screen options

satya164:
In React Navigation, the screen options can be specified statically. If you need to configure any options based on props and state of the component, or want to update state and props based on some action such as tab press, you need to do it in a hacky way by changing params. it's way more complicated than it needs to be. It also breaks when used with HOCs which don't hoist static props, a common source of confusion.

This PR adds a `setOptions` API to be able to update options directly without going through params.


example/index.tsx
```diff
@@ -65,20 +69,34 @@ const Second = ({
...
+}) => {
+  const [count, setCount] = React.useState(0);
+
+  React.useEffect(() => {
+    const timer = setInterval(() => setCount(c => c + 1), 1000);
+
+    return () => clearInterval(timer);
+  }, []);
+
+  useOptions<StackNavigationOptions>({
```
osdnk: is it really example, how options work? It's just a param

src/useDescriptors.tsx
```diff
@@ -69,11 +68,21 @@ export default function useDescriptors({
...
                 getState={getState}
                 setState={setState}
+                setOptions={setOptions}
               />
             </NavigationBuilderContext.Provider>
           );
         },
-        options: screen.options || EMPTY_OPTIONS,
+        options: {
+          ...(typeof screen.options === 'function'
```
osdnk: it's really nice to have it!

src/useDescriptors.tsx
```diff
@@ -69,11 +68,21 @@ export default function useDescriptors({
...
+            ? screen.options({
+                // @ts-ignore
```
osdnk: what's the issue here?

src/NavigationPropContext.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react';
+import { NavigationProp } from './types';
+
+const NavigationPropContext = React.createContext<NavigationProp | undefined>(
```
osdnk: It's rather unrelated change. Can we move it separate PR?

src/SceneView.tsx
```diff
@@ -28,8 +38,16 @@ export default function SceneView(props: Props) {
       setParams: (params: object, target?: TargetRoute<string>) => {
         helpers.setParams(params, target ? target : { key: route.key });
       },
+      setOptions: (options: object) =>
+        setOptions(o => ({
```
osdnk, line 42: AFAIK we don't want to have setOptions. If we can set it after mounting, why not to use param?

osdnk (approved):
I understand motivation of these changes and follow it. Can we move PropContext to separate diff? AFAIK we want to advise ppl not to use options if it's not very needed. What's the real use case of having options?

example/index.tsx
satya164: In this example, `author` is a param, but `title` is an option

src/NavigationPropContext.tsx
satya164: need to expose the context to access the navigation prop in the hook

src/useDescriptors.tsx
satya164: the type of `route` doesn't match

PR #21, opened 2019-07-21
feat: remove necessity of wrapping navigators into container

osdnk:


brentvatne:
the motivation for explicit containers was two-fold:
1) people were rendering navigators in nested components without realizing the implications of doing this - it would basically create a new navigation container with a navigation subtree that was disconnected from the parent, leading to some confusing behavior for people. in the new design on this repo this wouldn't be an issue because rendering navigators this way is one of the main motivations behind the design.
2) on web/native we want to use different navigation containers: https://github.com/react-navigation/web/blob/master/src/createBrowserApp.js and https://github.com/react-navigation/native/blob/master/src/createAppContainer.js. so, if we provide a default container then we either have to bias towards one of them or include both and pick automatically depending on platform.

osdnk:
1. It’s not an issue
2. Container is so far very small and designed for mobile. Assuming that most of react navigation users are focused on mobile and a lot of them does not need deeplinking etc, can’t we assume it will cover most of cases?

satya164:
The container also handles hardware back button on mobile which I think all apps which support Android need to handle. And for webs, I think history integration is something people will want to have. So users would need to wrap their code in container still.

Though we can still wrap by default with the base container without any additional functionality by default. When users use another container, that'll work fine. The only problem is that by not making it explicit, users may not realize that they need to do it to handle back button etc.

I don't have any strong opinion on this. So lemme know what you prefer.

osdnk:
@brentvatne it's up to you

src/__tests__/NavigationContainer.test.tsx
```diff
@@ -42,8 +45,12 @@ it('throws when setState is accessed without a container', () => {
   );
 });
 
-it('throws when performTransaction is accessed without a container', () => {
-  expect.assertions(1);
+it('throws when performTransaction is accessed without a container but inside navigator', () => {
```
satya164, line 48: doesn't throw?

src/__tests__/NavigationContainer.test.tsx
```diff
@@ -54,11 +61,11 @@ it('throws when performTransaction is accessed without a container', () => {
...
-  const element = <Test />;
-
-  expect(() => render(element).update(element)).toThrowError(
-    "We couldn't find a navigation context. Have you wrapped your app with 'NavigationContainer'?"
-  );
+  const element = (
+    <TestNavigator.Navigator>
+      <TestNavigator.Screen name="name" component={Test} />
+    </TestNavigator.Navigator>);
+  render(element).update(element);
```
satya164, line 68: nit: new line before xd

src/createNavigator.tsx
```diff
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import { ParamListBase, RouteConfig, TypedNavigator } from './types';
 import Screen from './Screen';
+import { NavigationStateContext } from './NavigationContainer';
+import { NavigationContainer } from './index';
```
satya164, line 5: move the import to `NavigationContainer` file like above to avoid circular dep

src/createNavigator.tsx
```diff
@@ -10,7 +12,21 @@ export default function createNavigator<N extends React.ComponentType<any>>(
     typeof RawNavigator
   > {
     return {
-      Navigator: RawNavigator,
+      Navigator: function Navigator(props) {
+        const { getState } = React.useContext(NavigationStateContext);
+
+        // @ts-ignore
+        const Navigator = <RawNavigator {...props} />
+        if (getState() === null) {
```
satya164, line 20: nit: new line before xd

src/createNavigator.tsx
satya164: nit: unrelated, but can you move the types import to bottom xd

satya164 (changes requested):

osdnk:
drob it

PR #22, opened 2019-07-21
feat: accept router to be a function

osdnk:


satya164:
Can you drop the commit regarding `setOptions` and rebase against master so it's easier to review?

osdnk:
@satya164 My changes are built on thew top of yours. If you're force-pushed commit with another hash, it might imply problems. I can do it after merging your PR.

src/useNavigationBuilder.tsx
```diff
@@ -49,10 +51,13 @@ const getRouteConfigsFromChildren = <ScreenOptions extends object>(
...
-export default function useNavigationBuilder<ScreenOptions extends object>(
-  router: Router<any>,
-  options: Options
-) {
+export default function useNavigationBuilder<
+  ScreenOptions extends object,
+  Options extends DefualtOptions = DefualtOptions
+>(router: RouterHelper<any, Options>, options: Options) {
+  const { current: currentRouter } = React.useRef<Router<any>>(
+    typeof router === 'function' ? router(options) : router
```
satya164: I think it's better to just always make it a function rather than accepting both object and function

satya164: I think we shouldn't pass things like `children` to the router. It'll be problematic if someone writing a custom router uses children.

src/useNavigationBuilder.tsx
```diff
@@ -79,7 +84,7 @@ export default function useNavigationBuilder<ScreenOptions extends object>(
   );
 
   const [initialState] = React.useState(() =>
-    router.getInitialState({
+    currentRouter.getInitialState({
       routeNames,
       initialRouteName,
```
satya164: Since we now pass `initialRouteName` as an option, we could remove it from here

src/useNavigationBuilder.tsx
```diff
@@ -79,7 +84,7 @@ export default function useNavigationBuilder<ScreenOptions extends object>(
...
       initialParamsList,
```
satya164: Let's rename this to `routeParamList`, since it's not really initial as it changes when children change.

src/useNavigationBuilder.tsx
```diff
@@ -49,10 +51,13 @@ const getRouteConfigsFromChildren = <ScreenOptions extends object>(
...
 
-export default function useNavigationBuilder<ScreenOptions extends object>(
-  router: Router<any>,
-  options: Options
-) {
+export default function useNavigationBuilder<
+  ScreenOptions extends object,
+  Options extends DefualtOptions = DefualtOptions
+>(router: RouterHelper<any, Options>, options: Options) {
+  const { current: currentRouter } = React.useRef<Router<any>>(
```
satya164: Since there's always going to be one router, I think no need to say `currentRouter`, `router` looks nicer 😛

src/useNavigationBuilder.tsx
```diff
@@ -49,10 +51,13 @@ const getRouteConfigsFromChildren = <ScreenOptions extends object>(
     );
   }, []);
 
-export default function useNavigationBuilder<ScreenOptions extends object>(
-  router: Router<any>,
-  options: Options
-) {
+export default function useNavigationBuilder<
+  ScreenOptions extends object,
+  Options extends DefualtOptions = DefualtOptions
```
satya164: `RouterOptions`

satya164 (changes requested):

src/types.tsx
```diff
@@ -55,6 +56,16 @@ export type ActionCreators<Action extends NavigationAction> = {
   [key: string]: (...args: any) => Action;
 };
 
+export type DefualtOptions = {
```
satya164: `RouterDefaultOptions`

src/types.tsx
```diff
@@ -104,14 +114,12 @@ export type Router<
    * @param state State object to update.
    * @param options.routeNames New list of route names.
    * @param options.initialRouteName Route to focus in the state.
-   * @param options.initialParamsList Object containing initial params for each route.
```
satya164, line 107: small typo

src/types.tsx
```diff
@@ -79,12 +91,10 @@ export type Router<
    *
    * @param options.routeNames List of valid route names as defined in the screen components.
    * @param options.initialRouteName Route to focus in the state.
-   * @param options.initialParamsList Object containing initial params for each route.
```
satya164, line 82: small typo

src/types.tsx
```diff
@@ -70,6 +71,17 @@ export type ActionCreators<Action extends NavigationAction> = {
   [key: string]: (...args: any) => Action;
 };
 
+export type DefaultRouterOptions = {
+  initialRouteName?: string;
+  children: React.ReactNode;
+};
+
+export type RouterHelper<
```
satya164: maybe `RouterCreator`?

src/types.tsx
```diff
@@ -70,6 +71,17 @@ export type ActionCreators<Action extends NavigationAction> = {
   [key: string]: (...args: any) => Action;
 };
 
+export type DefaultRouterOptions = {
+  initialRouteName?: string;
+  children: React.ReactNode;
```
satya164: router doesn't use children no?

satya164 (approved):

PR #25, opened 2019-07-22
refactor: only accept a key for setParams

satya164:
It doesn't make much sense for setParams to accept a name as you'd not want to set params for a random route.
So this commit changes the signature so that only a key can be passed.

It also adds a 'source' property to actions which the reducer can use if key wasn't provided.


osdnk (commented):
I'm not following these changes. For most cases (actually almost all cases I met) we use only one route of given name per app. It's truly uncommon that there's more than one route of given type and this change will imply increase of complexity of many apps.

E.g. route A exist only once in app and we want to set params from route B. The route B needs to know key of A (name is predictable, key not) which calls for increasing complexity of data workflow (developer need to pass key of A to B, which is unnecessary in most cases).

osdnk (approved):
meh

osdnk:
merch id

osdnk:
:shipit:

osdnk:
🚢 🇮🇹

osdnk:
🥖

PR #29, opened 2019-07-24
@osdnk/imperative dispatch

osdnk:


src/NavigationContainer.tsx
```diff
@@ -1,6 +1,14 @@
 import * as React from 'react';
 import EnsureSingleNavigator from './EnsureSingleNavigator';
-import { Route, NavigationState, InitialState, PartialState } from './types';
```
satya164, line 3: nit: types at the end :D

src/NavigationContainer.tsx
```diff
@@ -71,6 +78,32 @@ export default function NavigationContainer({
   const isTransactionActiveRef = React.useRef<boolean>(false);
   const isFirstMountRef = React.useRef<boolean>(true);
 
+  React.useImperativeHandle(ref, () => ({
+    dispatch: action => {
+      context.performTransaction(() => {
+        for (let i = 0; i < actionListeners.length; i++) {
+          actionListeners[i]({
+            type: '__DISPATCH_INTERNAL',
```
satya164: Instead of a special type, what do you think about changing the signature of `onAction` function? Something like:

```js
function onAction(action: NavigationAction, source?: { root?: boolean, target?: string, navigator?: string });
```

I think it's better because otherwise we're leaking internals as anyone can call `dispatch` with such an action.

src/__tests__/NavigationContainer.test.tsx
```diff
@@ -112,3 +121,132 @@ it('throws when nesting performTransaction', () => {
...
+    <NavigationContainer
+      ref={ref}
+      initialState={initialState}
+      onStateChange={onStateChange}
+    >
+      <ParentNavigator>
+        <Screen name="foo">{() => null}</Screen>
+        <Screen name="bar">{() => null}</Screen>
+        <Screen name="baz">
+          {() => (
```
satya164, line 222: Maybe we should add multiple navigators in screens to verify only the focused navigator handles this action (the focused screen could be the last one)

osdnk: yeah

satya164 (approved):

PR #33, opened 2019-07-28
feat: add focus and blur events

satya164:


benoitdion:
a bit outside the scope of this PR but what do you think about including AppState to focus/blur events? When using react-navigation we found that most of the time the work we'd do on focus/blur of a screen we'd also do when AppState changes. 

Related: https://github.com/facebook/react-native/issues/25127

satya164:
@benoitdion Maybe will be good to include when writing bindings for React Native in `useFocusEffect` https://github.com/react-navigation/navigation-ex/pull/35

Can you open an RFC here https://github.com/react-navigation/rfcs so we remember it?

PR #38, opened 2019-08-01
chore: setup monorepo with yarn workspaces

satya164:


packages/example/.gitignore
```diff
@@ -0,0 +1,11 @@
...
+.expo/*
+npm-debug.*
+*.jks
+*.p8
+*.p12
+*.key
+*.mobileprovision
+*.orig.*
+web-build/
+web-report/
```
thymikee: these 2 are featured in both gitignores, check which one you need. I think I'd rather go with a single gitignore in the root

packages/example/babel.config.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable import/no-commonjs */
+/* eslint-env node */
+
+module.exports = function(api) {
```
thymikee, line 4: There's supposed to be only one `babel.config.js` in a monorepo – in the root. But if it's working, then I guess it's ok. I'd try to change it to `.babelrc.js`

packages/example/metro.config.js
```diff
@@ -0,0 +1,46 @@
...
+            )}\\/.*$`
+          )
+      )
+    ),
+
+    providesModuleNodeModules: [
+      '@babel/runtime',
+      'react',
+      'react-native',
+      'shortid',
```
thymikee, line 34: make sure you need all of these, likely not

packages/example/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@navigation-ex/example",
+  "version": "0.0.0",
+  "private": true,
+  "workspaces": {
```
thymikee, line 5: workspaces is supposed to be present only in the root of the repo, isn't it?

packages/example/babel.config.js
satya164: Just kept what's generated by expo. Don't wanna mess with it 😅

packages/example/package.json
satya164: Yarn docs says we can use this to configure hoisting for specific packages. For the expo app, I just disabled hoisting since it causes issues with RN.

thymikee: Ok makes sense

thymikee (approved):

thymikee:
After this is done, feel free to setup lerna for less import mapping and easier publishing. The config is pretty simple with workspaces: https://github.com/react-native-community/cli/blob/master/lerna.json

PR #42, opened 2019-08-05
feat: integrate reanimated based stack

satya164:
ezgif-5-92c1a20cbab1

PR #43, opened 2019-08-05
 feat: add drawer navigator integration

satya164:


osdnk:
same https://github.com/react-navigation/navigation-ex/pull/42#pullrequestreview-271052259

packages/drawer/src/types.tsx
```diff
@@ -0,0 +1,117 @@
+import { StyleProp, ViewStyle, TextStyle } from 'react-native';
```
osdnk, line 1: everything for this point is copied right?

packages/drawer/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "@navigation-ex/core": "^0.0.1",
+    "@navigation-ex/routers": "^0.0.1",
+    "react-native-safe-area-view": "^0.14.6"
+  },
+  "devDependencies": {
+    "@types/react": "^16.8.24",
+    "@types/react-native": "^0.60.2",
+    "react": "16.8.3",
+    "react-native": "^0.59.8",
+    "react-native-gesture-handler": "^1.3.0",
```
osdnk, line 16: why it's dev dep?

osdnk (approved):
lovely!

packages/core/src/types.tsx
```diff
@@ -378,7 +395,7 @@ export type RouteConfig<
     | ScreenOptions
     | ((props: {
         route: RouteProp<ParamList, RouteName>;
-        navigation: NavigationHelpersCommon<ParamList>;
+        navigation: any;
```
osdnk: hm?

packages/drawer/package.json
satya164: basically so that tools like typescript and eslint will resolve it. similar to how we have it in tab view. in this monorepo, it'll be hoisted to top

packages/drawer/src/types.tsx
satya164: yeah, the only thing I have changed are types + replaced `onDrawerX` callbacks with events.

packages/core/src/types.tsx
satya164: screens receive a more complex navigation type (as in the README), and it's not really possible to type it properly here. that's why it's `any`. users need to type it manually like in README.

packages/drawer/src/types.tsx
satya164: oh, I also added `DrawerRouter` (second file from last)

PR #46, opened 2019-08-06
docs: add basic example app

satya164:
depends on #44


osdnk:
pushed to master after local rebasing, sorry

PR #48, opened 2019-08-08
feat: add native integration skeleton

osdnk:


packages/example/src/index.tsx
```diff
@@ -41,6 +46,8 @@ const PERSISTENCE_KEY = 'NAVIGATION_STATE';
 Asset.loadAsync(StackAssets);
 
 export default function App() {
+  const containerRef = React.useRef<NavigationContainerRef>(React.createRef());
```
satya164: You don't need to pass `createRef` into `useRef`, we can pass null and use the returned ref instead of `.current`

packages/core/src/NavigationContainer.tsx
```diff
@@ -162,7 +167,7 @@ const Container = React.forwardRef(function NavigationContainer(
         "Any 'setState' calls need to be done inside 'performTransaction'"
       );
     }
-
+    wasStateModified.current = true;
```
satya164: shouldn't we check if state actually changed?

packages/example/package.json
```diff
@@ -17,6 +17,7 @@
     "eject": "expo eject"
   },
   "dependencies": {
+    "@navigation-ex/native": "^0.0.1",
```
satya164: I think this will create a symlink, right? if just add it to the eslint config in root

satya164 (dismissed):

packages/native/src/useBackButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+export default function useBackButton(ref: NavigationContainerRef) {
+  React.useEffect(() => {
+    BackHandler.addEventListener('hardwareBackPress', () => {
+      const castedRef = ref as React.RefObject<
+        NavigationHelpers<ParamListBase>
+      >;
+      if (castedRef.current !== null) {
+        castedRef.current
+          .dispatch({ type: 'GO_BACK' })
+          .then((wasHandled: boolean) => {
```
satya164: Use async/await maybe :)

satya164:
I have another thought. This wouldn't be a problem for us if we had a some kind of `navigation.canGoBack` method to query if we can go back in history. Each router can implement this method.

When you call this method, it'll recursively call parent `canGoBack` like the `isFocused` method does, and in the end, we have concrete info on if back button can be handled.

Similar to how the root container's `dispatch` dispatches the action to focused navigator, root container's `canGoBack` will call the focused navigator's `canGoBack`.

Now we can check this method in our container and only dispatch a back action if we can go back. No messing with how state updates, no returning promise from dispatch etc. `canGoBack` is also a useful method for users.

packages/example/src/index.tsx
```diff
@@ -49,6 +54,8 @@ const PERSISTENCE_KEY = 'NAVIGATION_STATE';
...
+  useNativeIntegration(containerRef.current);
   const [isReady, setIsReady] = React.useState(false);
```
satya164, line 64: nit: newline before

packages/native/src/index.tsx
```diff
@@ -0,0 +1,8 @@
+import { NavigationContainerRef } from '@navigation-ex/core';
+import useBackButton from './useBackButton';
+
+export { useBackButton };
+
+export default function useNativeIntegration(ref: NavigationContainerRef) {
```
satya164: move the hook to separate file since we only re-export in `index` files everywhere

packages/core/src/NavigationContainer.tsx
```diff
@@ -86,18 +87,25 @@ const getPartialState = (
...
+  } = useFocusedListeners();
+
   const dispatch = (action: NavigationAction) =>
-    context.performTransaction(() => {
-      actionListeners[0](action, undefined, null);
-    });
+    listeners[0](navigation => navigation.dispatch(action));
 
   React.useImperativeHandle(ref, () => ({
+    performOnFocusedNavigator: (
```
satya164: Don't think we want to expose `performOnFocusedNavigator`. The exposed ref should look like every other navigation object. We should expose `canGoBack` method instead where we internally perform this on focused navigator and return the result.

satya164 (changes requested):
It's a bit hard to review with both changes in this PR. Can you rebase your PR and squash the commits so that there is a single commit for changes made only in this PR which I can easily review?

Or let's get the other PR merged first and rebase.

PR #50, opened 2019-08-08
feat: add canGoBack

osdnk:


packages/core/src/useNavigationHelpers.tsx
```diff
@@ -68,6 +68,9 @@ export default function useNavigationHelpers<Action extends NavigationAction>({
       isFocused: parentNavigationHelpers
         ? parentNavigationHelpers.isFocused
         : () => true,
+      canGoBack: parentNavigationHelpers
```
satya164: maybe we also need to consider if current navigator can go back? it's different from `isFocused` cause `isFocused` checks if the navigator is focused, `canGoBack` checks if navigator can handle `goBack`.

so it could be something like:

```js
const canGoBack =
  router.canGoBack(getState()) ||
  parentNavigationHelpers
    ? parentNavigationHelpers.canGoBack()
    : false;
```

packages/core/src/useNavigationCache.tsx
```diff
@@ -74,6 +83,13 @@ export default function useNavigationCache<
             return navigation ? navigation.isFocused() : true;
           },
           isFirstRouteInParent: () => isFirst,
+          canGoBack() {
```
satya164: if we consider router state in the `canGoBack` of navigation helpers, we don't need to implement it here :)

satya164: note: due to caching, this value will also be incorrect when state changes, so better to move it to navigation helpers

packages/core/src/useNavigationCache.tsx
```diff
@@ -74,6 +83,13 @@ export default function useNavigationCache<
             return navigation ? navigation.isFocused() : true;
           },
           isFirstRouteInParent: () => isFirst,
```
satya164: maybe we should remove this since we have `canGoBack`? not totally sure

packages/core/src/types.tsx
```diff
@@ -154,6 +154,13 @@ export type Router<
    */
   shouldActionChangeFocus(action: NavigationAction): boolean;
 
+  /**
+   * Whether the back action will be handled by navigation
+   *
+   * @param state State object to apply the action on.
```
satya164:
```suggestion
   * @param state State object to check.
```

packages/core/src/types.tsx
```diff
@@ -264,6 +271,12 @@ type NavigationHelpersCommon<
    * To get notified of focus changes, use `addListener('focus', cb)` and `addListener('blur', cb)`.
    */
   isFocused(): boolean;
+
+  /**
+   * Check if dispatching back action will be handled by navigation.
+   * Note that this method doesn't re-render screen result changes. So don't u`se it in `render`.
```
satya164: if we implement it like I mentioned, I think it'll work with re-renders :)

packages/core/src/BaseRouter.tsx
```diff
@@ -78,6 +78,10 @@ const BaseRouter = {
   shouldActionChangeFocus(action: CommonAction) {
     return action.type === 'NAVIGATE';
   },
+
+  canGoBack() {
+    return false;
```
satya164, line 83: I think we shouldn't add it here. we should implement this property in tab router instead

packages/core/src/__tests__/__fixtures__/MockRouter.tsx
```diff
@@ -98,6 +98,9 @@ export default function MockRouter(options: DefaultRouterOptions) {
     },
 
     actionCreators: {},
```
satya164: nit: guess this `actionCreators` can be removed

packages/core/src/__tests__/useDescriptors.test.tsx
```diff
@@ -54,6 +57,180 @@ it('sets options with options prop as an object', () => {
       `);
 });
 
+it('handles properly canGoBack with no canGoBack overrode', () => {
```
satya164:
```suggestion
it("returns correct value for canGoBack when it's not overridden", () => {
```

packages/core/src/__tests__/useDescriptors.test.tsx
```diff
@@ -54,6 +57,180 @@ it('sets options with options prop as an object', () => {
...
+      </TestNavigator>
+    </NavigationContainer>
+  );
+
+  render(root).update(root);
+
+  expect(result).toEqual(false);
+});
+
+it('handles properly canGoBack with canGoBack overrode', () => {
```
satya164:
```suggestion
it("returns correct value for canGoBack when it's overridden", () => {
```

packages/core/src/__tests__/useDescriptors.test.tsx
```diff
@@ -54,6 +57,180 @@ it('sets options with options prop as an object', () => {
...
+      </ParentNavigator>
+    </NavigationContainer>
+  );
+
+  render(root).update(root);
+
+  expect(result).toEqual(true);
+});
+
+it('handles properly canGoBack with canGoBack overrode but not as parent', () => {
```
satya164:
```suggestion
it('returns correct value for canGoBack when parent router overrides it', () => {
```

packages/core/src/types.tsx
satya164: on another thought, maybe leave it. re-renders might trigger for a lot of things

packages/core/src/useNavigationCache.tsx
osdnk: not sure. Maybe in follow up

packages/core/src/BaseRouter.tsx
osdnk: yes, but I'll still leave it by default for now

satya164: what I'm afraid of is that by leaving it here it's easy for routers to forget to implement it. since it's quite crucial to proper back button handling, maybe be it's better to force routers to implement it?

packages/core/src/NavigationContainer.tsx
```diff
@@ -88,9 +88,7 @@ const Container = React.forwardRef(function NavigationContainer(
 
   const dispatch = (action: NavigationAction) =>
     context.performTransaction(() => {
-      for (let i = 0; i < actionListeners.length; i++) {
```
satya164, line 91: why this change?

satya164: oh I get it, cause there can only be one navigator, right?

packages/core/src/useNavigationBuilder.tsx
```diff
@@ -198,6 +198,7 @@ export default function useNavigationBuilder<
     getState,
     setState,
     emitter,
+    router,
```
satya164, line 201: unneeded?

packages/core/src/types.tsx
satya164:
```suggestion
   * Note that this method doesn't re-render screen when the result changes. So don't use it in `render`.
```

satya164 (approved):
LGTM, just few minor comments

packages/core/src/NavigationContainer.tsx
satya164: I'd keep the loop. the container shouldn't know that each navigator always adds only one listener

osdnk: why? I see it needed bc we will need to return this value soon while wanted to implement back behavior

packages/core/src/useNavigationBuilder.tsx
osdnk: needed

osdnk:
```
      canGoBack: () =>
        router.canGoBack(getState()) ||
        (parentNavigationHelpers && parentNavigationHelpers.canGoBack()) ||
        false,
```

PR #51, opened 2019-08-09
fix: implement canGoBack for tab router

satya164:


packages/routers/src/TabRouter.tsx
```diff
@@ -211,6 +211,10 @@ export default function TabRouter({
       return action.type === 'NAVIGATE';
     },
 
+    canGoBack(state) {
+      return router.getStateForAction(state, { type: 'GO_BACK' }) !== null;
```
satya164, line 215: this gave me an idea. we don't actually need to implement this method in routers. we can instead do this to implement `canGoBack`:

```js
canGoBack: () =>
        router.getStateForAction(getState(), BaseActions.goBack()) !== null ||
        (parentNavigationHelpers && parentNavigationHelpers.canGoBack()) ||
        false
```

osdnk: idk.

osdnk (approved):

PR #53, opened 2019-08-13
refactor: simplify dispatching logic

osdnk:


packages/core/src/useNavigationHelpers.tsx
```diff
@@ -82,4 +94,27 @@ export default function useNavigationHelpers<
     setState,
     onAction,
   ]);
+
+  const { addNavigationListener } = React.useContext(NavigationFocusedContext);
```
satya164: I think it's better to move this code to a separate hook. `useNavigationHelpers` is to create navigation helper object, so it's not the right place for this. Maybe `useFocusedListeners`

packages/core/src/NavigationFocusedContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+export type FocusedNavigationCallback = (
+  navigation: NavigationHelpers<ParamListBase>
+) => any;
+
+export type FocusedNavigationListener = (
+  callback: FocusedNavigationCallback
+) => { handled: boolean; result: any };
+
+const MISSING_CONTEXT_ERROR =
+  "We couldn't find a navigation focused context. Have you wrapped your app with 'NavigationContainer'?";
```
satya164: Maybe we should't throw this error? A "navigation focused context" is an internal thing, so the error message could be confusing for the user

packages/core/src/NavigationFocusedContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Context which holds the required helpers needed to access focused navigator.
+ */
+
+const NavigationFocusedContext = React.createContext<{
```
satya164: I've removed the need for a `removeXListener`, so maybe we can use the same builder context for this now?

packages/core/src/NavigationFocusedContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+    throw new Error(MISSING_CONTEXT_ERROR);
+  },
+});
+
+export default NavigationFocusedContext;
+
+/**
+ * Hook which lets child navigators add action listeners.
+ */
+export function useFocusedListeners() {
```
satya164: Let's move this hook to separate file like other hooks. We could move the listener logic you added in `useNavigationHelpers` into this hook because they are related logic.

packages/core/src/NavigationFocusedContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+  "We couldn't find a navigation focused context. Have you wrapped your app with 'NavigationContainer'?";
+
+/**
+ * Context which holds the required helpers needed to access focused navigator.
+ */
+
+const NavigationFocusedContext = React.createContext<{
+  addNavigationListener: (listener: FocusedNavigationListener) => void;
+}>({
+  get addNavigationListener(): any {
```
satya164: Not sure about `addNavigationListener` since we're not listening to navigation. Maybe something like `addFocusedListener` to match the naming of the context and the hook?

packages/core/src/useNavigationBuilder.tsx
```diff
@@ -18,6 +18,7 @@ import {
   Router,
   RouterFactory,
 } from './types';
+import { useFocusedListeners } from './NavigationFocusedContext';
```
satya164: nit: types at the end :D

packages/core/src/useNavigationHelpers.tsx
```diff
@@ -11,6 +11,10 @@ import {
   ParamListBase,
   Router,
 } from './types';
+import NavigationFocusedContext, {
```
satya164: nit: types at the end

packages/core/src/useNavigationHelpers.tsx
```diff
@@ -21,6 +25,7 @@ type Options<State extends NavigationState, Action extends NavigationAction> = {
   setState: (state: State) => void;
   emitter: NavigationEventEmitter;
   router: Router<State, Action>;
+  navigationListeners: FocusedNavigationListener[];
```
satya164: Perhaps `focusedListeners` following naming of the hook and the context?

packages/core/src/useNavigationHelpers.tsx
```diff
@@ -82,4 +94,27 @@ export default function useNavigationHelpers<
     setState,
     onAction,
   ]);
+
+  const { addNavigationListener } = React.useContext(NavigationFocusedContext);
+  const listener = React.useRef((callback: FocusedNavigationCallback) => {
+    if (navigation.isFocused()) {
```
satya164: I was wondering if the `target` approach we following in `onAction` is better than checking `navigation.isFocused` as `isFocused` will check focused state of each parent. but lets leave for separate PR

packages/core/src/NavigationContainer.tsx
```diff
@@ -92,10 +93,10 @@ const Container = React.forwardRef(function NavigationContainer(
     getPartialState(initialState)
   );
 
+  const { listeners, addNavigationListener } = useFocusedListeners();
+
   const dispatch = (action: NavigationAction) =>
-    context.performTransaction(() => {
-      actionListeners[0](action, undefined, null);
-    });
+    listeners[0](navigation => navigation.dispatch(action));
```
satya164, line 99: Still think there's no reason not to loop over the listeners.

packages/core/src/NavigationFocusedContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+  callback: FocusedNavigationCallback
+) => { handled: boolean; result: any };
+
+const MISSING_CONTEXT_ERROR =
+  "We couldn't find a navigation focused context. Have you wrapped your app with 'NavigationContainer'?";
+
+/**
+ * Context which holds the required helpers needed to access focused navigator.
+ */
+
```
satya164: I think there shouldn't be new line here or JSDoc hint won't work

packages/core/src/useNavigationHelpers.tsx
```diff
@@ -82,4 +94,27 @@ export default function useNavigationHelpers<
     setState,
     onAction,
   ]);
+
+  const { addNavigationListener } = React.useContext(NavigationFocusedContext);
+  const listener = React.useRef((callback: FocusedNavigationCallback) => {
```
satya164: Why `useRef` here? Maybe you wanted to use `useCallback`? Using ref is incorrect since the values will never update.

packages/core/src/useNavigationHelpers.tsx
```diff
@@ -82,4 +94,27 @@ export default function useNavigationHelpers<
...
+      return { handled: true, result: callback(navigation) };
+    } else {
+      return { handled: false, result: null };
+    }
+  });
+
+  React.useEffect(() => {
+    const { current: currentListener } = listener;
+    return addNavigationListener(currentListener);
+  });
```
satya164: Need to add a dependency array to avoid the effect running every render.

satya164 (changes requested):

packages/core/src/NavigationContainer.tsx
satya164: Should wrap this in curly braces so the return value of `dispatch` isn't a boolean

packages/core/src/__tests__/useFocusEffect.test.tsx
```diff
@@ -45,17 +45,23 @@ it('runs focus effect on focus change', () => {
...
+  act(() => {
+    navigation.current.navigate('second');
+  });
 
   expect(focusEffect).toBeCalledTimes(1);
   expect(focusEffectCleanup).not.toBeCalled();
 
-  act(() => navigation.current.navigate('third'));
+  act(() => {
+    navigation.current.navigate('third');
```
satya164: I think we should revert this change because it caught an issue where we were leaking internal return value as the return value of `dispatch`

satya164 (changes requested):

packages/core/src/NavigationFocusedContext.tsx
osdnk: I think that it's not important. I just didn't want to make keys optional.

packages/core/src/NavigationFocusedContext.tsx
osdnk: can you clarify?

osdnk: Oh you mean to construct useEffect directly here. It makes sense, but I don't have opinion if it's obviously better

packages/core/src/useNavigationHelpers.tsx
osdnk: no strong opinion

packages/core/src/NavigationContainer.tsx
osdnk: I think since we're absolutely sure that here cannot be more than one listener it's ok

packages/core/src/NavigationFocusedContext.tsx
osdnk: got it! thanks!

packages/core/src/NavigationFocusedContext.tsx
osdnk: maybe I can change default to: `() => {}`

osdnk: Or whatever, it can be optional. Fixed!

PR #54, opened 2019-08-14
refactor: simpler way to check canGoBack

satya164:


osdnk (approved):
I think it should be somehow documented, but ok 👍

PR #57, opened 2019-08-15
Initial work on deep linking and web support

satya164:
Please rebase and merge instead of squash when merging :)

I have made rehydration more robust to add all missing fields and handle invalid state more gracefully as we discussed before. This means you can pass as minimum of data as needed as initial state.

For example:

```js
{
  routes: [
    {
      name: 'foo',
      state: {
        routes: [
          {
            name: 'bar',
            params: { count: 42 },
          },
        ],
      },
    },
  ],
}
```

I have also added 2 helper utilities for converting between URL and state. Converting is lossy because it's not really possible to keep all data in the URL (e.g. params for non-terminal route). But this kind of data can be stored in the history object with `history.pushState` etc.

I'm not very happy with the utilities currently because of the way it handles params. Currently it'll `JSON.stringify` the params which works, but looks ugly in the URL. However we can think of how to improve it later.


packages/core/src/getStateFromPath.tsx
```diff
@@ -0,0 +1,50 @@
...
+    } else {
+      result = state;
+    }
+
+    current = state;
+    segments.shift();
+  }
+
+  if (current == null || result == null) {
+    throw new Error(`Unable to parse path to state: ${path}`);
```
osdnk: I'm not sure if we want to throw an exception here. It's not problem of our code, that given url is broken. Maybe we should fallback to initial state?

packages/core/src/getStateFromPath.tsx
```diff
@@ -0,0 +1,50 @@
...
+ */
+export default function getStateFromPath(path: string): InitialState {
+  const parts = path.split('?');
+  const segments = parts[0].split('/').filter(Boolean);
+  const query = parts[1] ? parts[1].split('&') : undefined;
+
+  let result: InitialState | undefined;
+  let current: InitialState | undefined;
+
+  while (segments.length) {
```
osdnk, line 19: Can't we make it with reduce?

packages/routers/src/StackRouter.tsx
```diff
@@ -47,36 +48,69 @@ export default function StackRouter(options: StackRouterOptions) {
...
         key: `stack-${shortid()}`,
-        index,
+        index: 0,
         routeNames,
-        routes: routeNames.slice(0, index + 1).map(name => ({
-          name,
-          key: `${name}-${shortid()}`,
-          params: routeParamList[name],
-        })),
+        routes: [
```
osdnk, line 60: makes sense, but AFAIK it's unrelated? Can you confirm?

packages/core/src/getPathFromState.tsx
```diff
@@ -0,0 +1,47 @@
+import { NavigationState, PartialState, Route } from './types';
+
+type State = NavigationState | Omit<PartialState<NavigationState>, 'stale'>;
+
+/**
+ * Utility to serialize a navigation state object to a path string.
+ * This is a lossy operation because we can't store all of the data in the path.
```
osdnk: This comment maybe is not needed? `Utility to serialize a navigation state object to a path string.` should be fine

packages/core/src/__tests__/getStateFromPath.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+              params: { count: 42 },
+            },
+          ],
+        },
+      },
+    ],
+  });
+});
+
+it('handles ending slash when converting', () => {
```
osdnk, line 55: 😍

packages/core/src/getStateFromPath.tsx
satya164: We can, but I feel it's simpler with a loop.

packages/routers/src/StackRouter.tsx
satya164: yeah, not strictly related

PR #58, opened 2019-08-17
feat: add integration with redux devtools extension

satya164:
Currently supports:
- Tracking actions and navigation state
- Time travel for navigation state

It doesn't do anything in production


osdnk (approved):
looks awesome 👍

PR #60, opened 2019-08-21
chore: configure lerna with bob

satya164:


README.md
```diff
@@ -451,10 +451,10 @@ Unfortunately it's not possible to verify that the type of children elements are
 
 ## Contributing
 
-The project uses a monorepo structure for the packages managed by [yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/). To get started with the project, run `yarn` to install the required dependencies for each package:
+The project uses a monorepo structure for the packages managed by [yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) and [lerna](https://lerna.js.org). To get started with the project, run `lerna bootstrap` to install the required dependencies for each package:
 
 ```sh
-yarn
+lerna bootstrap
```
thymikee, line 457: AFAIK with `"useWorkspaces": true` you don't have to do `lerna bootstrap`, `yarn install` will do just fine

lerna.json
```diff
@@ -0,0 +1,8 @@
+{
+  "version": "0.0.0",
```
thymikee, line 2: should be synchronized with latest package version, so "0.0.1" seems more suitable currently

packages/bottom-tabs/package.json
```diff
@@ -1,21 +1,43 @@
...
+  "react-native": "src/index.tsx",
+  "module": "lib/module/index.js",
+  "types": "lib/typescript/src/index.d.ts",
+  "files": [
+    "src",
+    "lib"
+  ],
+  "scripts": {
+    "prepare": "bob build",
+    "clean": "del lib"
```
thymikee, line 15: Do you need local scripts at all? It's a common thing to run everything from the root dir. I guess this is a workaround to make `bob` work for every package?

packages/core/src/useNavigationBuilder.tsx
```diff
@@ -20,8 +20,13 @@ import {
   Router,
   RouterFactory,
   PartialState,
+  PrivateValueStore,
 } from './types';
 
+// This is to make TypeScript compiler happy
+// eslint-disable-next-line babel/no-unused-expressions
+PrivateValueStore;
```
thymikee, line 28: Is this some kind of a circular dep? Maybe mark as a TODO to fix later

packages/drawer/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "../../tsconfig"
```
thymikee, line 2: I see that you build per package now. But if you ever wanted to build from a single entry point, you'll likely need ["references"](https://www.typescriptlang.org/docs/handbook/project-references.html). This will help tsc figure out the order of resolving local deps and will speed up builds. See example: https://github.com/facebook/jest/blob/bc5c3c7c49f8b190caf060319cc2d88255d9b2ff/packages/jest-cli/tsconfig.json#L7-L14

tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "compilerOptions": {
     "baseUrl": ".",
     "paths": {
-      "@navigation-ex": ["./packages"]
+      "@navigation-ex/*": ["./packages/*/src", "./packages/*/lib/typescript"],
```
thymikee, line 5: with "references" you likely wouldn't need absolute paths

thymikee (approved):
This is a bit different than what I'm used to, but provided that you want to use Bob on per-package basis and also Metro handles source files (so you don't need a watch mode in development to compile files), this is fine 👍

packages/core/src/useNavigationBuilder.tsx
satya164: Nope not circular. Probably TS complains because this type is not exported and all properties on it are private.

PR #61, opened 2019-08-21
feat: compatibility layer

satya164:
**TODO**
- [x] Events (`focus` -> `willFocus`, `blur` -> `willBlur`, `transitionEnd` -> `isFocused` ? `didFocus` : `didBlur`)
- [x] Action signatures
- [x] Merge `Screen.navigationOptions` with the one in route config
- [x] Add warnings for unsupported usage


packages/compat/README.md
```diff
@@ -0,0 +1,28 @@
...
+yarn add @react-navigation/core @react-navigation/compat
+```
+
+## Usage
+
+```js
+import { createCompatNavigatorFactory } from '@react-navigation/compat';
+import { createStackNavigator } from '@react-navigation/stack';
+
+const RootStack = createCompatNavigatorFactory(createStackNavigator)(
```
osdnk, line 19: I'm not following this naming. It suggests that we're creating a factory, which is not true. I'd rather call it somehow simpler. Maybe `createCompatNavigator` or `createCompat`?

osdnk (approved):
looks good 🤷‍♂

packages/compat/README.md
satya164: > It suggests that we're creating a factory, which is not true

Are we not? `createCompatNavigatorFactory(createStackNavigator)` returns a factory function (say `createCompatStackNavigator`), which you then call with the options.

PR #62, opened 2019-08-21
feat: add dangerouslyGetParent

osdnk:


packages/core/src/useNavigationCache.tsx
```diff
@@ -87,6 +91,13 @@ export default function useNavigationCache<
           ...rest,
           ...helpers,
           ...emitter.create(route.key),
+          dangerouslyGetParent: () =>
+            parentNavigation
```
satya164: I think this should be `navigation`, not `parentNavigation`. Reading the React navigation docs, this refers to the navigation object of the navigator.

PR #69, opened 2019-08-24
fix: blinking header in stack

osdnk:
Previously descriptor stopped existing while animating back. 
It's okay, then header was undefined. 
So `hasCustomHeader` was changing which was implying blinking. 

Now if no descriptor, we're assuming the recent prop when descriptor was existing


osdnk:
closing in favor on fix directly on master

PR #83, opened 2019-08-28
feat: handle navigating with both with both key and name

osdnk:


packages/core/src/CommonActions.tsx
```diff
@@ -47,12 +48,9 @@ export function navigate(...args: any): Action {
   } else {
     const payload = args[0];
 
-    if (
-      (payload.hasOwnProperty('key') && payload.hasOwnProperty('name')) ||
-      (!payload.hasOwnProperty('key') && !payload.hasOwnProperty('name'))
-    ) {
+    if (!payload.hasOwnProperty('key') && !payload.hasOwnProperty('name')) {
```
satya164, line 54: I think we also need update the signature of the first overload

satya164 (commented):
LGTM. Can you also update the signature of the `navigate` function in `NavigationProp` types?

satya164 (approved):

PR #84, opened 2019-08-29
fix: change interpolated style when idle to avoid messing up reanimated

satya164:


osdnk (approved):
https://www.youtube.com/watch?v=8ELbX5CMomE

PR #94, opened 2019-09-02
feat: integrate `InteractionManager` in stack

janicduplessis:
In RN Animated `InteractionManager` is integrated automatically when starting animations but this is not the case in reanimated. This integrates it for stack.


satya164:
9563a27284b9be7c84380ad44c8c2cd9defcee8e

PR #97, opened 2019-09-13
feat: make deep link handling more flexible

satya164:
This adds ability to specify a custom config to control how to convert between state and path.

Example:

```js
{
  Chat: {
    path: 'chat/:author/:id',
    parse: { id: Number }
  }
}
```

The above config can parse a path matching the provided pattern: `chat/jane/42` to a valid state:

```js
{
  routes: [
    {
      name: 'Chat',
      params: { author: 'jane', id: 42 },
    },
  ],
}
```

This makes it much easier to control the parsing without having to specify a custom function.


jineshshah36:
What if you need multiple paths to match the same route name?

brentvatne:
@jineshshah36 - ~can you explain? this doesn't seem to make sense. if two route names mapped to '/hello' then which route definition would you use?~

edit: my bad i read that wrong, it's a valid question

jineshshah36:
To clarify, I would expect something like the following:

```ts
{
  Chat: [
    {
      path: "chat/:author/:id",
      parse: { id: Number },
    },
    {
      path: "some/:other/:path",
      parse: { path: Number },
    },
  ],
}
```

satya164:
Sounds reasonable. Maybe for a followup PR

satya164:
Sounds reasonable. Maybe for a followup PR.

One thing to keep in mind is that in the current PR, this works both ways: you can convert between a state and a path using the same config. If multiple paths can represent the same route, then this mapping isn't 2-way anymore, i.e. we can convert a path to a navigation state, but not the other way around. Perhaps when 2 configs specified, we should treat the first one as the primary config, which will be used for converting state to path.

Why 2-way conversion is needed? Mainly because otherwise we wouldn't be able to support web (I think).

jineshshah36:
That makes sense. I would say the primary reason to have multiple routes match a path is to make it easier to maintain compatibilty when someone changes a path. So, having the first or a “default” marked path take precedence makes sense.

I’d imagine this would be even more important when supporting web since links can’t always all be rewritten if they are on other sites and such. Alternatively, a redirect mechanism in the config could also work.

package.json
```diff
@@ -31,7 +31,7 @@
     "@types/jest": "^24.0.13",
     "codecov": "^3.5.0",
     "commitlint": "^8.0.0",
-    "core-js": "^3.1.4",
+    "core-js": "^3.2.1",
```
osdnk, line 34: Can you tell why we need these changes?

osdnk (approved):
I'm happy with these changes. If you're feeling good with this API, feel free to merge!

package.json
satya164: @osdnk jest was giving some issue. this is only for tests so not very important.

PR #100, opened 2019-09-16
fix: don't remove route if animation isn't finished when dragging

osdnk:
The problem was that if we're starting dragging after triggering the animation (it might happen spontaneously when we're making a super-fast move) the route was removed from route array. It's not expected.


packages/stack/src/views/Stack/Card.tsx
```diff
@@ -574,7 +574,13 @@ export default class Card extends React.Component<Props> {
                 this.offset,
                 cond(
                   this.distance,
-                  divide(this.gesture, this.distance),
+                  divide(
+                    multiply(
+                      this.gestureUntraversed,
```
osdnk, line 579: This is vital because of some race condition. I'm not fully confident, but this inlining cannot have any negative impact and fix flickering. Flickering became an issue after fix from the previous commit.

packages/stack/src/views/Stack/Card.tsx
```diff
@@ -574,7 +574,13 @@ export default class Card extends React.Component<Props> {
...
+                      I18nManager.isRTL ? MINUS_ONE_NODE : TRUE_NODE
```
satya164, line 580: Instead of `TRUE_NODE`, can we do `PLUS_ONE_NODE`?

osdnk: yes we can but since I made this decision in many places I want to do it in follow up

PR #104, opened 2019-09-20
refactor(stack): convert Header component to function

dulmandakh:
Just converted to functional component. I find it easy to read and comprehend.


packages/stack/src/views/Header/Header.tsx
```diff
@@ -4,64 +4,56 @@ import HeaderSegment from './HeaderSegment';
...
-    } = this.props;
-    const { options } = scene.descriptor;
-    const title =
-      typeof options.headerTitle !== 'function' &&
-      options.headerTitle !== undefined
-        ? options.headerTitle
-        : options.title !== undefined
-        ? options.title
-        : scene.route.name;
+export default React.memo(function(props: HeaderProps) {
```
osdnk: Can you explain why you use `React.memo` here? AFAIK there’s no need for it

osdnk (commented):
Looks good!

packages/stack/src/views/Header/Header.tsx
dulmandakh: previously it was extending PureComponent, which is exact same as React.memo.

osdnk: Yes, but following my understanding it should work well w/o this wrapper

dulmandakh: yes, it's kind of a performance optimization. React.memo will shallow compare props, and won't re-render if nothing changed.

dulmandakh: it's like a performance optimization provided by PureComponent

dulmandakh:
@satya164  @osdnk please review and merge

packages/stack/src/views/Header/Header.tsx
satya164: We need to give the component a name so that it shows up in devtools:

```suggestion
export default React.memo(function Header(props: HeaderProps) {
```

satya164 (changes requested):

packages/stack/src/views/Header/Header.tsx
dulmandakh: done 👍

satya164 (approved):

PR #105, opened 2019-09-20
feat(stack): use Animated.Text for header title

dulmandakh:
This PR changes HeaderTitle to use Animated.Text, so that developers can use Animated values, like opacity, in headerTitleStyle.


dulmandakh:
IIRC, react navigation 4.x and below supported Animated values in headerTitleStyle.

dulmandakh:
I use it to show title if a content is scrolled to a certain point

dulmandakh:
@satya164 please review and merge

packages/stack/src/views/Header/HeaderTitle.tsx
```diff
@@ -1,12 +1,12 @@
...
-import { Text, StyleSheet, Platform, TextProps } from 'react-native';
+import { Animated, StyleSheet, Platform, TextProps } from 'react-native';
 
 type Props = TextProps & {
   children?: string;
 };
 
 export default function HeaderTitle({ style, ...rest }: Props) {
-  return <Text {...rest} style={[styles.title, style]} />;
+  return <Animated.Text {...rest} style={[styles.title, style]} />;
```
satya164, line 10: We're using Reanimated in stack. So we should change this to import `Animated` from `react-native-reanimated` instead.

satya164 (changes requested):

packages/stack/src/views/Header/HeaderTitle.tsx
dulmandakh: done 👍

satya164 (approved):

PR #106, opened 2019-09-20
feat(stack): use Animated.View for header background

dulmandakh:
This PR changes HeaderBackground to use Animated.View, so that it will support Animated.Value in style.


dulmandakh:
@satya164 please review and merge

packages/stack/src/views/Header/HeaderBackground.tsx
```diff
@@ -1,8 +1,8 @@
 import * as React from 'react';
-import { View, StyleSheet, Platform, ViewProps } from 'react-native';
+import { Animated, StyleSheet, Platform, ViewProps } from 'react-native';
 
 export default function HeaderBackground({ style, ...rest }: ViewProps) {
-  return <View style={[styles.container, style]} {...rest} />;
+  return <Animated.View style={[styles.container, style]} {...rest} />;
```
satya164, line 6: We're using Reanimated in stack. So we should change this to import `Animated` from `react-native-reanimated` instead.

satya164 (changes requested):

satya164 (approved):

PR #107, opened 2019-09-20
Add include in rule

jpstrikesback:
Hey this little change allowed me to load the example app (web) (albeit with nothing else working :D), Still something up with tsconfig (I think?)


satya164:
Thanks for the PR! Will check it.

PR #109, opened 2019-09-23
feat: export BottomTabBar props type

dulmandakh:
I customize bottom tab bar wrapping BottomTabBar, and I need its props type. I'm not sure if it's proper way to do it, but Props from views is final type used in the component.


osdnk (approved):
Thanks!

Issue #111, opened 2019-09-24
Crash when reset from nested navigator


leethree:
### Current Behavior

The following exception is thrown when tries to `reset` app to top-level from nested navigator:
```
TypeError: undefined is not an object (evaluating 'screen.options')

Stack trace:
  /<redacted>/navigation-ex/packages/core/src/useDescriptors.tsx:130:30 in options
  /<redacted>/navigation-ex/packages/stack/src/views/Stack/StackView.tsx:91:23 in isAnimationEnabled
  /<redacted>/navigation-ex/packages/stack/src/views/Stack/StackView.tsx:100:29 in
  ...
```

### Expected Behavior

App should be reset to the destination state.

### How to reproduce

- Add the `reset` code to the example app:
```patch
diff --git a/packages/example/src/Screens/SimpleStack.tsx b/packages/example/src/Screens/SimpleStack.tsx
index 7d16651..05dc8a0 100644
--- a/packages/example/src/Screens/SimpleStack.tsx
+++ b/packages/example/src/Screens/SimpleStack.tsx
@@ -27,7 +27,10 @@ const ArticleScreen = ({
     <View style={styles.buttons}>
       <Button
         mode="contained"
-        onPress={() => navigation.push('album')}
+        onPress={() => navigation.reset({
+          index: 0,
+          routes: [{ name: 'compat-api', key: 'whatever' }],
+        })}
         style={styles.button}
       >
         Push album

```
- Open the example app, go to "Simple Stack" or "Bottom Tabs".
- Tap "Push Album"
- Crash 

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 76de32f441653c9866d6194fa324f1e31ed1f5b3
| react-native     | 0.59.10
| expo             | ^34.0.1
| node             | 10.15.0
| npm or yarn      | yarn 1.17.3

satya164:
The `reset` method resets the state of the current navigator. Looks like you're trying to update the state of a parent navigator. In which case, you need to specify which navigator's state you want to change using the `target` key:

```js
navigation.dispatch({
  ...CommonActions.reset({
    index: 0,
    routes: [{ name: 'compat-api', key: 'whatever' }],
  }),
  target: 'state.key-of-the-navigator'
});
```

leethree:
Ah I see. It's not ideal because the screens would need to know about how the navigation is structured. 

[...]

PR #112, opened 2019-09-24
feat: add a method to reset root navigator state

satya164:
Context https://github.com/react-navigation/navigation-ex/issues/111#issuecomment-534625379

Not sure if this is a good idea.


osdnk (approved):
I'm not sure about concept, but code looks good. Let's merge!

PR #119, opened 2019-09-30
feat: add a getRootState method

osdnk:


packages/core/src/NavigationBuilderContext.tsx
```diff
@@ -14,6 +19,7 @@ export type FocusedNavigationListener = <T>(
   callback: FocusedNavigationCallback<T>
 ) => { handled: boolean; result: T };
 
+export type NavigatorStateGetter = () => NavigationState;
```
satya164, line 22: nit: newline after

packages/core/src/NavigationBuilderContext.tsx
```diff
@@ -25,6 +31,7 @@ const NavigationBuilderContext = React.createContext<{
   addActionListener?: (listener: ChildActionListener) => void;
   addFocusedListener?: (listener: FocusedNavigationListener) => void;
   onRouteFocus?: (key: string) => void;
+  addStateGetter?: (getter: NavigatorStateGetter, key: string) => void;
```
satya164: nit: change signature to `(key, getter)` like `key, value`

packages/core/src/NavigationContainer.tsx
```diff
@@ -15,6 +15,7 @@ import {
   NavigationContainerRef,
   NavigationContainerProps,
 } from './types';
+import useStateGetters from './useStateGetters';
```
satya164: nit: move before types

packages/core/src/NavigationContainer.tsx
```diff
@@ -134,6 +137,10 @@ const Container = React.forwardRef(function NavigationContainer(
     [trackAction]
   );
 
+  const getRootState = () => {
+    return Object.values(stateGetters.current)[0]();
```
satya164: Instead of getting the first value, can we be more explicit and get `stateGetters.current.root()`?

packages/core/src/__tests__/NavigationContainer.test.tsx
```diff
@@ -321,3 +321,71 @@ it('handle resetting state with ref', () => {
...
 });
+
+it('handle getRootState', () => {
+  const TestNavigator = (props: any) => {
+    const { state, descriptors } = useNavigationBuilder(MockRouter, props);
+
+    return descriptors[state.routes[state.index].key].render();
+  };
+
+  const initialState = {
```
satya164: Maybe lets remove `initialState`. Just makes the test bigger and is not needed

packages/core/src/useNavigationBuilder.tsx
```diff
@@ -254,6 +258,16 @@ export default function useNavigationBuilder<
     focusedListeners,
   });
 
+  const getStateForRoute = React.useCallback(
```
satya164: Move this to `useStateGetters` hook, and return this instead of exposing the `stateGetters` ref

packages/core/src/useOnGetState.tsx
```diff
@@ -0,0 +1,31 @@
...
+  getStateForRoute: (routeName: string) => NavigationState;
+  getState: () => NavigationState;
+}) {
+  const { addStateGetter } = React.useContext(NavigationBuilderContext);
+  const route = React.useContext(NavigationRouteContext);
+  const key = route ? route.key : 'root';
+
+  const getter = React.useCallback(
+    () => ({
+      ...getState(),
```
satya164: maybe call `getState` once?

```
const state = getState();

// ...

return {
  ...state
```

packages/core/src/useOnGetState.tsx
```diff
@@ -0,0 +1,31 @@
...
+      routes: getState().routes.map(route => ({
+        ...route,
+        state: route.state ? getStateForRoute(route.key) : route.state,
```
satya164: `route.state` might be undefined, but there can still be a child navigator. Change to:

```suggestion
        state: getStateForRoute(route.key),
```

We'll need to change `getStateForRoute` to handle non-existent keys and return `undefined`

packages/core/src/useOnGetState.tsx
```diff
@@ -0,0 +1,31 @@
...
+      })),
+    }),
+    [getState, getStateForRoute]
+  );
+
+  React.useEffect(() => {
+    addStateGetter && addStateGetter(getter, key);
```
satya164: Need to return for the cleanup function to take effect

packages/core/src/useStateGetters.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import { NavigatorStateGetter } from './NavigationBuilderContext';
+
+/**
+ * Hook which lets child navigators add getters to be called for obtaining stage.
```
satya164:
```suggestion
 * Hook which lets child navigators add getters to be called for obtaining rehydrated state.
```

packages/core/src/useStateGetters.tsx
```diff
@@ -0,0 +1,25 @@
...
+ */
+export default function useStateGetters() {
+  const stateGetters = React.useRef<Record<string, NavigatorStateGetter>>({});
+
+  const addStateGetter = React.useCallback(
+    (getter: NavigatorStateGetter, key: string) => {
+      stateGetters.current[key] = getter;
+
+      return () => {
+        delete stateGetters.current[key];
```
satya164: Avoid `delete`. It de-opts in JS VMs. Set to `undefined` instead
```suggestion
        stateGetters.current[key] = undefined;
```

packages/core/src/useStateGetters.tsx
```diff
@@ -0,0 +1,25 @@
...
+      };
+    },
+    [stateGetters]
```
satya164: This shouldn't need to be in the deps array.
```suggestion
    []
```

satya164 (approved):

PR #127, opened 2019-10-16
fix: don't fade incoming background when fading header

leethree:
During a cross-fade, the header on the back should still have `opacity=1` to prevent the header become translucent and avoid anything in the back showing up.

The problem with the current animation is that opacities of layers are multiplied instead of added. Two white view with `opacity=0.5` on top of each other isn't equivalent to one white view with `opacity=1`.

opacity

### DIFF
--- a/packages/stack/src/TransitionConfigs/HeaderStyleInterpolators.tsx
+++ b/packages/stack/src/TransitionConfigs/HeaderStyleInterpolators.tsx
@@ -101,8 +101,8 @@ export function forFade({
 }: StackHeaderInterpolationProps): StackHeaderInterpolatedStyle {
   const progress = add(current.progress, next ? next.progress : 0);
   const opacity = interpolate(progress, {
-    inputRange: [0, 1, 2],
-    outputRange: [0, 1, 0],
+    inputRange: [0, 1, 1.9999, 2],
+    outputRange: [0, 1, 1, 0],
   });
 
   return {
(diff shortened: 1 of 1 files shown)


packages/stack/src/TransitionConfigs/HeaderStyleInterpolators.tsx
```diff
@@ -101,8 +101,8 @@ export function forFade({
 }: StackHeaderInterpolationProps): StackHeaderInterpolatedStyle {
   const progress = add(current.progress, next ? next.progress : 0);
   const opacity = interpolate(progress, {
-    inputRange: [0, 1, 2],
-    outputRange: [0, 1, 0],
+    inputRange: [0, 1, 1.9999, 2],
```
satya164: What's the reason to have `1.9999` instead of changing `0` to `1` in the `outputRange`?

satya164 (changes requested):
Thanks for the PR. I think changing `backgroundStyle: { opacity }` to `backgroundStyle: { opacity: current.progress }` would achieve the effect we're looking for while still cross-fading the text. Wdyt?

packages/stack/src/TransitionConfigs/HeaderStyleInterpolators.tsx
leethree: It is to hide the screen when the transition is finished. I can remove it if it's not necessary.

leethree:
@satya164 yeah it should achieve the same thing but I feel `interpolate` is more generic and easier to understand.

satya164:
It's not the same thing, changing background opacity only means the text still cross-fades. Also `progress` is a value ranging from `0` to `1` so interpolation is unnecessary for opacity.

leethree:
@satya164 I see what you mean. I updated the code.

satya164 (approved):

PR #145, opened 2019-10-30
fix: drop isFirstRouteInParent method

satya164:
The `isFirstRouteInParent` method was added to determine whether the back button should be shown in the header for stack navigator or not.
This was mainly due to the API of the old version of stack whose public API of header didn't have all required info to determine whether it should be shown.
It was probably a mistake to add it, because this method doesn't look at history and so pretty much useless for other navigators which aren't stack.

In the new stack, the header receives a `previous` prop and the `headerLeft` option receives a `canGoBack` prop which can be used for this purpose.
It's also possible to check if a route is the first by doing `navigation.dangerouslyGetState().routes[0].key === route.key`.

So it's time to drop this method.


osdnk (approved):
Nice catch!

osdnk (approved):

PR #151, opened 2019-10-31
fix: minor tweaks for web and fix example

satya164:


osdnk (approved):

packages/example/package.json
```diff
@@ -29,6 +29,7 @@
     "react-native-gesture-handler": "~1.3.0",
     "react-native-paper": "^3.1.1",
     "react-native-reanimated": "~1.2.0",
```
osdnk, line 31: what happened here?

PR #153, opened 2019-10-31
feat: add headerBackTitleVisible to navigation options in native stack

phamfoo:
https://github.com/react-navigation/navigation-ex/issues/132


satya164 (approved):
Thanks for the PR

PR #158, opened 2019-11-02
feat: support transform style for header

Freddy03h:
Two related things in this PR:

- add `transform` in the `safeStyles` headerStyle
- add the `safeStyles` headerStyle as argument in the `headerBackground` function

Use case: **Create a collapsible scrollview header**. To migrate this use case from 4 to 5 this PR is needed but the philosophy isn't far from previous react-navigation. You need to set `headerTransparent` to `true` and use a custom `headerBackground` (even if don't really want a custom one).

It's needed to transform the `headerBackground` and the View that wrap the headerLeft/Right and the headerTitle.
Also, giving `headerStyle` as argument to `headerBackground` function make it easier to customize by setting an `headerStyle` option in a screen instead of recreate an entire `headerBackground` for each screen.

### DIFF
--- a/packages/stack/src/types.tsx
+++ b/packages/stack/src/types.tsx
@@ -185,7 +185,9 @@ export type StackHeaderOptions = {
    * This is useful for using backgrounds such as an image or a gradient.
    * You can use this with `headerTransparent` to render a blur view, for example, to create a translucent header.
    */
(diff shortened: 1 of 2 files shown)


satya164 (approved):

Freddy03h:
thank's for the approval

PR #159, opened 2019-11-03
Allow passing headerTitle as a render function

jgking2:
Allow ability to pass a custom header Component


satya164:
Thanks for the PR. This was not exposed because it wasn't working as expected:

- On Android, setting headerTitle as a React element replaces the title, but the custom text appears on the right half of the header, where title would've ended, not on the left like you'd expect it to.
- On iOS, setting headerTitle as a React element removes the back button :O

jgking2:
@satya164 definitely ran across that, I'm seeing the android issue, but on iOS it seems to be working ok (may not be looking at the specific case).  I can definitely see why it's not exposed though!

osdnk:

PR #161, opened 2019-11-04
fix: fix default BottomTabBar button

oliverwinter-vdb:
Hey, 

I noticed that creating a new BottomTabNavigator required the TabBarButton to be specifically defined by the user, and then saw that the default fallback for renderbutton would never resolve as it is always a function (returning the user defined button, or undefined). This small fix moves the default fallback to the BottomTabView, so a user can use the default navigation tab bar button.


satya164 (approved):
Oops! Thanks

PR #167, opened 2019-11-10
refactor: rename createNavigator since it doesn't create a navigator

satya164:


josheleonard:
@satya164 
Could you please release @react-navigation/bottom-tabs 5.0.0-alpha.23 since 5.0.0-alpha.22 crashes due to it internally calling `createNavigator` instead of `createNavigatorFactory`

satya164:
@josheleonard update @react-navigation/core

josheleonard:
Ok I double checked my stuff, I was on the latest /core but my monorepo had a package that used an older version of /bottom-tabs. I updated my /bottom-tabs dependency version in that package to the latest and now everything is working as expected. Thank you for your help.

PR #170, opened 2019-11-13
fix: pass labelStyle prop in DrawerItem label

ddedic:
In current version, labelStyle prop is never passed to the DrawerItem label component.
This pull request simply passes that prop to the Text component.


satya164 (approved):
Thanks for the PR

PR #171, opened 2019-11-13
chore: migrate to Github Actions

satya164:


satya164:
Bummer. Not possible to run tasks in parallel with shared jobs.

PR #173, opened 2019-11-14
feat: support headerLeft for native-stack

janicduplessis:
Same thing as `headerRight`. This is useful to customize the back button to use `Cancel` text for example.

### DIFF
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -91,6 +91,10 @@ export type NativeStackNavigationOptions = {
    */
   headerLargeTitle?: boolean;
   /**
+   * Function which returns a React Element to display on the left side of the header.
+   */
+  headerLeft?: () => React.ReactNode;
+  /**
    * Function which returns a React Element to display on the right side of the header.
    */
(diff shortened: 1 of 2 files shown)


satya164:
Thanks for the PR. The reason it's not exposed is basically the same reason as https://github.com/react-navigation/navigation-ex/pull/159#issuecomment-549106370

On Android, `headerLeft` will remove both the title and back button. So it doesn't work like user would expect it to. I think we need to re-create the title and put it manually to make it work on Android.

janicduplessis:
@satya164 Do you think this should be fixed in rn-screens instead?

satya164:
@janicduplessis yeah. I think better to keep the integration a small wrapper and address issues in screens itself.

janicduplessis:
Should we merge this then? :D

I can have a look at the issue when testing my app on Android

satya164:
I'd rather wait for it to be fixed in screens first.

radko93:
This is now fixed https://github.com/kmagiera/react-native-screens/commit/0919404b2e71839fc551c5c61f7001b3622ffd92

osdnk:

PR #177, opened 2019-11-19
fix: allow passing partial params to `setParams`

Titozzz:
Since params get merged, no need to send them all I think :)


slorber (approved):
LGTM

satya164 (approved):
Thanks

PR #199, opened 2019-12-09
refactor: use proc node to improve reanimated mount time

satya164:
Co-authored-by: Michał Osadnik <micosa97@gmail.com>


osdnk:
Does it really halp?

satya164:
Don't think so, but I think it improves the code anyway by extracting out animation stuff, so good to have it anyway.

PR #206, opened 2019-12-12
fix: use Partial type for initialParam

oliverwinter-vdb:
Based on this commit https://github.com/react-navigation/navigation-ex/commit/11efb066429a3fc8b7e8e48d897286208d9a5449

Does it make sense that you can set initialParams as a partial of the routeParams as they are now merged with the pushed screen?


satya164 (approved):
Makes sense. Thanks

PR #212, opened 2019-12-15
chore: update README expo and native-stack

eriveltonelias:
Expo SDK 36 now support react-native-screens v2, expo users can use native-stack package.


satya164 (approved):
Thanks

PR #237, opened 2020-01-03
fix: interaction manager in stack

osdnk:


packages/stack/src/views/Stack/Card.tsx
```diff
@@ -159,12 +159,15 @@ export default class Card extends React.Component<Props> {
       spec.animation === 'spring' ? Animated.spring : Animated.timing;
 
     onTransitionStart?.({ closing: Boolean(closing) });
+    this.handleStartInteraction();
```
satya164, line 161: small thing: can we set it before `onTransitionStart`? so user can still rely on it in `transitionStart` event

PR #244, opened 2020-01-04
fix: change transition spec ios for spring animation

Jekiwijaya:
fix: change transition spec ios for spring animation so finish callback could be trigger smoothly
https://github.com/facebook/react-native/issues/20783#issuecomment-481372109

Current value make animated finished callback has ~300ms delayed, this effect
https://github.com/react-navigation/navigation-ex/blob/master/packages/stack/src/views/Stack/Card.tsx#L169

### DIFF
--- a/packages/stack/src/TransitionConfigs/TransitionSpecs.tsx
+++ b/packages/stack/src/TransitionConfigs/TransitionSpecs.tsx
@@ -11,8 +11,8 @@ export const TransitionIOSSpec: TransitionSpec = {
     damping: 500,
     mass: 3,
     overshootClamping: true,
-    restDisplacementThreshold: 0.01,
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Thanks for the PR.

The current animation config is aimed to replicate native iOS animations. I think the proposed changes don't feel right. The difference in the feel of the animation is quite noticeable, there is kind of snap towards the end of the animation which we don't want.

> so finish callback could be trigger smoothly

Wdym by callback triggering smoothly?

> Current value make animated finished callback has ~300ms delayed, this effect

Can you elaborate what's the issue with the linked line?

satya164:
Closing due to no response

PR #269, opened 2020-01-16
feat: support for callbackAction on useScrollToTop hook

eriveltonelias:
in some cases after scrolled to top will be awesome to refresh the page.

### DIFF
--- a/packages/native/src/useScrollToTop.tsx
+++ b/packages/native/src/useScrollToTop.tsx
@@ -43,7 +43,10 @@ function getScrollableNode(ref: React.RefObject<ScrollableWrapper>) {
 }
 
 export default function useScrollToTop(
-  ref: React.RefObject<ScrollableWrapper>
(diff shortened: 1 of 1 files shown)


satya164:
Thanks for the PR.

I'm not convinced that we should support this. If you want to refetch data on scroll to top, wouldn't it be better to listen to `onScroll` and do it?

Another issue is that with such a callback, I'd expect that it'd fire after scroll to top ends. But it's not supported.

Note that it's also possible to achieve what you want with something like the following:

```js
function Albums() {
  const ref = React.useRef(null);

  useScrollToTop(
    React.useRef({
      scrollToTop() {
        // Trigger scroll to top for the component
        ref.scrollTo({ y: 0, animated: true });

        // Custom code
      },
    })
  );

  return <ScrollView ref={ref}>{/* content */}</ScrollView>;
}
```

eriveltonelias:
That's make sense, thanks! 

I'll add this code on our docs.

PR #271, opened 2020-01-16
feat: pass screenOptions to Screen's options function

benevbright:
I like to get `screenOptions` on `<Stack.Screen>`'s `options` function. 

What do you think? 
If you like this idea, I will continue to finish PR properly.

### DIFF
--- a/packages/core/src/useDescriptors.tsx
+++ b/packages/core/src/useDescriptors.tsx
@@ -105,6 +105,13 @@ export default function useDescriptors<
     (acc, route) => {
       const screen = screens[route.name];
       const navigation = navigations[route.key];
+      const screenOptionsObj = (typeof screenOptions === 'object' || screenOptions == null
+      ? screenOptions
+      : screenOptions({
+          // @ts-ignore
+          route,
+          navigation,
+        }))
 
       acc[route.key] = {
         navigation,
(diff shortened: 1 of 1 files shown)


satya164:
Couldn't you put `screenOptions` in a variable and access it?

```js
function App() {
  const screenOptions = {
    // whatever
  };

  return (
    <Stack.Navigator screenOptions={screenOptions}>
      <Stack.Screen
        name="Home"
        component={Home}
        options={
          {
            // you can access `screenOptions` here
          }
        }
      />
    </Stack.Navigator>
  );
}
```

Not convinced that we need a special API for it.

benevbright:
@satya164 Yes, of course, you can do it.
But I'm writing a library(https://github.com/benevbright/react-navigation-collapsible/tree/v5) that injecting the transformed `options`. In this case, I have to ask users to put `screenOptions` as well, which is inconvenient.

But it's just an idea. If you don't think it's relevant, please let me know. I will close it.

satya164:
In the stack package we don't do such merging of styles automatically, so it's probably better not to do such merging in a library. We'll revisit it in future, but I think currently it's better not to add it.

benevbright:
cool. It makes sense. Thanks for the response. 👍

PR #273, opened 2020-01-18
feat: add useNavigationState hook

satya164:
Sometimes it's useful to get the current navigation state inside a screen. We have the `dangerouslyGetState` method for that. However, the problem with this method is that it won't trigger a re-render when it changes, so user cannot rely on it for rendering something.

This adds a 2 things:
1. A `state` event similar to `focus` and `blur` that user can subscribe to
2. A `useNavigationState` hook that takes a selector and returns part of the state

Internally `useNavigationState` subscribes to the state event to get the current navigation state.

I have also made it mandatory to pass a selector to `useNavigationState`. This makes it harder to accidentally get the whole navigation state, which will trigger a re-render every time anything changes, even if we don't care about the change. With a selector, we can tell which part we care about, and if that part didn't change, it won't trigger a re-render.

For example, to get the same functionality as the old `isFirstRouteInParent` method:

```js
function MyComponent({ route }) {
  const isFirstRouteInParent = useNavigationState(state => state.routes[0] === route);

  // content
}
```


jeniasaigak:
Hi @satya164. Thanks for this PR it is really helpful but there is one issue with it and I want to help you prevent future issues.
With current implementation if the selector change dynamically the result will not change until the next state change so the value won't match provided selector.
So here is my proposition (it might be a little bit messy but the idea should be clear):
```javascript
function useNavigationState<T>(selector: Selector<T>): T {
  const navigation = useNavigation();
  const activeSelectorRef = useRef(selector);

  const [result, setResult] = React.useState(() =>
    selector(navigation.dangerouslyGetState())
  );

  // Recalculate result if selector has changed
  React.useEffect(() => {
    if (activeSelectorRef.current !== selector) {
      setResult(selector(navigation.dangerouslyGetState()));
      activeSelectorRef.current = selector;
    }
  }, [navigation, selector]);

  React.useEffect(() => {
    return navigation.addListener("state", e => {
      setResult(activeSelectorRef.current(e.data.state));
    });
  }, [navigation]);

  return result;
}
```
And here is how it should be used:
```javascript
useNavigationState(React.useCallback(state => state.whatever, []))
// OR
useNavigationState(staticSelectorFunction)
```

I understand that my usage example looks worse than `useNavigationState(state => state.whatever)` but it should work more right and more in "React way" because it will not lead to inappropriate value

P.S. Sorry if the code have some errors: I didn't try it, just wanted to share the idea.

satya164:
@zmefz thanks for your comment. I think the simplest fix is to just re-run the selector on re-render so we always select the value from the latest state. I just pushed a commit doing it.

PR #275, opened 2020-01-21
feat: Block pointer events if overlay is enabled

MichielDeMey:
This prevents pointer events from bleeding through the overlay which would result in background components being interactable when the overlay is active.

Great if you've created your own `TransitionPreset` with a custom `cardStyleInterpolator`.


satya164:
Hey! Thanks for the PR, but I'm not sure about this.

The problem is that when you have a spring animation, it takes a long time for the screen (along with the overlay) to be removed, which would mean that the previous screen won't be intractable until the screen fully dismisses (even if you no longer see it), which isn't what you'd expect.

For your use case, I think placing a `View` which fills entire screen and intercepts the touches is a good enough solution.

osdnk:

PR #279, opened 2020-01-22
feat: check for state type

NoemiRozpara:
Fulfills the task: `Add warning for non-serializable values in navigation state`

Navigation state is checked against undefined or object. In a case of different value, gives a warning. 

Note: 
needed to add`__DEV__` global to the `package.json` - tests failing otherwise.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,9 @@
     ],
     "moduleNameMapper": {
       "@react-navigation/([^/]+)": "<rootDir>/packages/$1/src"
+    },
+    "globals": {
+      "__DEV__": true
     }
   },
   "prettier": {
--- a/packages/core/src/NavigationContainer.tsx
+++ b/packages/core/src/NavigationContainer.tsx
@@ -1,3 +1,5 @@
+/* global __DEV__ */
+
 import * as React from 'react';
 import * as CommonActions from './CommonActions';
 import EnsureSingleNavigator from './EnsureSingleNavigator';
@@ -242,6 +244,15 @@ const Container = React.forwardRef(function NavigationContainer(
   );
 
   React.useEffect(() => {
+    if (__DEV__) {
+      if (
+        typeof state !== 'undefined' &&
+        Object.prototype.toString.call(state) !== '[object Object]'
+      ) {
+        console.warn('Navigation state contains non-serialiable value');
+      }
+    }
+
     if (skipTrackingRef.current) {
       skipTrackingRef.current = false;
     } else {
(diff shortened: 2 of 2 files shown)


packages/core/src/NavigationContainer.tsx
```diff
@@ -242,6 +242,16 @@ const Container = React.forwardRef(function NavigationContainer(
   );
 
   React.useEffect(() => {
+    // eslint-disable-next-line no-undef
+    if (__DEV__) {
```
satya164, line 246: Let's use 
```suggestion
    if (process.env.NODE_ENV !== 'production') {
```

packages/core/src/NavigationContainer.tsx
```diff
@@ -242,6 +242,16 @@ const Container = React.forwardRef(function NavigationContainer(
...
+      if (
```
satya164, line 247: We'll also need to traverse through the properties in the object and check each property if it's serializable.

Here is an implementation that you can copy: https://github.com/expo/ex-navigation/blob/ec8fa74dc3e94e2386ff6d344892d549d2407185/src/ExNavigationRouter.js#L308

But replace:

```js
_.isUndefined(obj) -> obj === undefined
_.isNull(obj) -> obj === null
_.isBoolean(obj) -> typeof obj ==='boolean'
_.isNumber(obj) -> typeof obj ==='number'
_.isString(obj) -> typeof obj ==='string'
_.isPlainObject(obj) -> typeof obj === 'object' && Object.prototype.toString.call(obj) !== '[object Object]'
_.isArray(obj) -> Array.isArray(obj)
```

package.json
```diff
@@ -64,6 +64,9 @@
     ],
     "moduleNameMapper": {
       "@react-navigation/([^/]+)": "<rootDir>/packages/$1/src"
+    },
+    "globals": {
```
satya164, line 68: We can use `process.env` and remove this

satya164 (changes requested):

packages/core/src/NavigationContainer.tsx
```diff
@@ -242,6 +242,16 @@ const Container = React.forwardRef(function NavigationContainer(
...
+        typeof state !== 'undefined' &&
+        Object.prototype.toString.call(state) !== '[object Object]'
+      ) {
+        console.warn('Navigation state contains non-serialiable value');
```
satya164, line 251:
```suggestion
        console.warn("We found non-serializable values in the navigation state, which can break usage such as persisting and restoring state. This might happen if you passed a function or other non-serializable values in params. If you need to use functions in your options, you can use 'navigation.setOptions' instead.");
```

satya164 (changes requested):

satya164:
https://github.com/react-navigation/navigation-ex/pull/283

PR #284, opened 2020-01-24
refactor: split Overlay into a new component

EvanBacon:
- Disable touch highlight on mobile Safari

### DIFF
--- a/packages/drawer/src/index.tsx
+++ b/packages/drawer/src/index.tsx
@@ -11,7 +11,6 @@ export { default as DrawerItem } from './views/DrawerItem';
 export { default as DrawerItemList } from './views/DrawerItemList';
 export { default as DrawerContent } from './views/DrawerContent';
 export { default as DrawerContentScrollView } from './views/DrawerContentScrollView';
-export { default as Overlay } from './views/Overlay';
 
 /**
  * Utilities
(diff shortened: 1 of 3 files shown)


packages/drawer/src/views/Overlay.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+const { interpolate, cond, greaterThan } = Animated;
+
+const PROGRESS_EPSILON = 0.05;
+
+type Props = React.ComponentProps<typeof Animated.View> & {
+  progress: Animated.Node<number>;
+};
+
+const Overlay = React.forwardRef(function Overlay(
```
satya164, line 13: Do we need the `forwardRef` wrapper?

EvanBacon: This works better with gestures on web

satya164: You mean gesture handler needs it?

packages/drawer/src/index.tsx
```diff
@@ -11,6 +11,7 @@ export { default as DrawerItem } from './views/DrawerItem';
 export { default as DrawerItemList } from './views/DrawerItemList';
 export { default as DrawerContent } from './views/DrawerContent';
 export { default as DrawerContentScrollView } from './views/DrawerContentScrollView';
+export { default as Overlay } from './views/Overlay';
```
satya164: I think we don't need to export it for the public API. Can you remove it? Then I'll merge

```suggestion
```

satya164 (approved):

PR #293, opened 2020-01-24
feat: add parsing empty path string

WoLewicki:
Added parsing empty string in deep linking. It is made by adding the `initial` property to the config route, which is the initial route for the app. In getPathFromState, if `initial` property occurs in the config, an empty string is returned immediately. In getStateFromPath, if the given path is an empty string, we look for `initial` property in the config and return state object referring to the given config.

### DIFF
--- a/packages/core/src/__tests__/getPathFromState.test.tsx
+++ b/packages/core/src/__tests__/getPathFromState.test.tsx
@@ -292,7 +292,7 @@ it('handles nested object with stringify in it', () => {
   };
 
   expect(getPathFromState(state, config)).toBe(path);
-  expect(getPathFromState(getStateFromPath(path))).toBe(path);
+  expect(getPathFromState(getStateFromPath(path, config), config)).toBe(path);
 });
 
 it('handles nested object for second route depth', () => {
@@ -326,7 +326,7 @@ it('handles nested object for second route depth', () => {
   };
 
   expect(getPathFromState(state, config)).toBe(path);
-  expect(getPathFromState(getStateFromPath(path))).toBe(path);
+  expect(getPathFromState(getStateFromPath(path), config)).toBe(path);
 });
 
 it('handles nested object for second route depth and and path and stringify in roots', () => {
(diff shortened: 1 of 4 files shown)


packages/core/src/getPathFromState.tsx
```diff
@@ -60,14 +64,18 @@ export default function getPathFromState(
         pattern = currentOptions[route.name] as string;
         break;
       } else if (typeof currentOptions[route.name] === 'object') {
+        // if there is `initial` prop in config, we should return empty string immidiately
```
osdnk:
```suggestion
        // if there is `initial` prop in the config, we should return empty string immediately
```

packages/core/src/getStateFromPath.tsx
```diff
@@ -206,10 +229,11 @@ function createNormalizedConfigs(
             value.parse ? (value.parse as ParseConfig) : undefined
           )
         );
-      } else if (nestedKey === 'parse') {
+      } else if (nestedKey === 'parse' || nestedKey === 'initial') {
         // We handle custom parse function when a `path` is specified (in nestedKey === path)
+        // `initial` prop should be ignored if path is not empty string
```
osdnk:
```suggestion
        // `initial` prop should be ignored if a path is not the empty string
```

packages/core/src/getStateFromPath.tsx
```diff
@@ -206,10 +229,11 @@ function createNormalizedConfigs(
...
       } else {
-        // If the name of the key is not `path` or `parse`, it's a nested config for route
+        // If the name of the key is not `path` or `parse` or `initial`, it's a nested config for route
```
osdnk:
```suggestion
        // If the name of the key is not `path` or `parse` or `initial`, it's a nested config for the route
```

packages/core/src/__tests__/getPathFromState.test.tsx
```diff
@@ -351,6 +482,15 @@ it('handles nested object for second route depth and and path and stringify in r
...
 
+  const path = '';
+
+  expect(getPathFromState(state, pathConfig)).toBe(path);
+  expect(
+    getPathFromState(getStateFromPath(path, stateConfig), pathConfig)
+  ).toBe(path);
+});
+
+it('converts state with initial prop in nested config to empty path string', () => {
```
osdnk:
```suggestion
it('converts state with initial prop in nested config to the empty path string', () => {
```

PR #294, opened 2020-01-24
fix: screens integration on Android

osdnk:


packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -128,6 +130,38 @@ function CardContainer({
...
+    ({ value }: { value: number }) => {
+      const newIsUnderEpsilon = value <= EPSILON ? 'box-none' : 'none';
+      if (newIsUnderEpsilon !== pointerEvents) {
+        setPointerEvents(newIsUnderEpsilon);
+      }
+    },
+    [pointerEvents]
+  );
+
+  const listenerRef = React.useRef<string>();
```
satya164: You don't need a ref. Move it to the callback to `useEffect`, then you won't need to set it to undefined either.

packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -128,6 +130,38 @@ function CardContainer({
 
   const { colors } = useTheme();
 
+  const [pointerEvents, setPointerEvents] = React.useState<'box-none' | 'none'>(
+    'box-none'
+  );
+
+  const valueListenerCallback = React.useCallback(
```
satya164: You can also move the callback to inside `useEffect`. It'll remove the need for adding `useCallback`.

packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -128,6 +130,38 @@ function CardContainer({
...
+    ({ value }: { value: number }) => {
+      const newIsUnderEpsilon = value <= EPSILON ? 'box-none' : 'none';
+      if (newIsUnderEpsilon !== pointerEvents) {
```
satya164: You don't need to check, afaik React will skip the update if the value is the same (only for `useState` hook). BTW it's better to use the updater callback to check previous state.

packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -128,6 +130,38 @@ function CardContainer({
 
   const { colors } = useTheme();
 
+  const [pointerEvents, setPointerEvents] = React.useState<'box-none' | 'none'>(
+    'box-none'
+  );
+
+  const valueListenerCallback = React.useCallback(
+    ({ value }: { value: number }) => {
+      const newIsUnderEpsilon = value <= EPSILON ? 'box-none' : 'none';
```
satya164: Maybe different variable name?

satya164 (changes requested):

packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -128,6 +130,38 @@ function CardContainer({
...
+    },
+    [pointerEvents]
+  );
+
+  const listenerRef = React.useRef<string>();
+
+  React.useEffect(() => {
+    if (scene.progress.next) {
+      // @ts-ignore
+      listenerRef.current = scene.progress.next.addListener(
```
satya164: Use optional chaning

```suggestion
      listenerRef.current = scene.progress.next?.addListener(
```

packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -128,6 +130,38 @@ function CardContainer({
...
+        valueListenerCallback
+      );
+    }
+    return () => {
+      if (scene.progress.next) {
+        // @ts-ignore
+        scene.progress.next.removeListener(listenerRef.current);
```
satya164: Use optional chaning

```suggestion
        scene.progress.next?.removeListener(listenerRef.current);
```

satya164 (approved):

satya164 (approved):
Thanks a ton for this contribution!
I'm extremely thankful for these changes and proud of these great lines of code!

I strongly encourage you for further development and can't wait for the next great changes! Do not give up doing a good job!

😍

kopax:
Hello, thanks for your contribution, unfortunatly, this break my migration to v5 final :

```jsx
 @react-navigation/drawer  ^5.0.0-alpha.41  →        ^5.0.0 
 @react-navigation/native  ^5.0.0-alpha.22  →        ^5.0.0 
 @react-navigation/stack   ^5.0.0-alpha.63  →        ^5.0.0 
```

https://github.com/react-navigation/react-navigation/issues/363

error is in there: https://github.com/react-navigation/react-navigation/blob/250fa9abc2174e39b80fe976428044ab4869abc1/packages/stack/src/views/Stack/CardStack.tsx#L42 

All my test now are failing with:

```
    TypeError: Cannot read property 'split' of undefined

      at Object.<anonymous> (node_modules/@react-navigation/stack/lib/commonjs/views/Stack/CardStack.tsx:42:26)
      at Object.<anonymous> (node_modules/@react-navigation/stack/lib/commonjs/views/Stack/StackView.tsx:8:1)
```

not sure how to do now, should I mock this ? It could break all my test environment... Thanks!

kopax:
Without 

```
const maybeExpoVersion = global.Expo?.Constants.manifest.sdkVersion.split(
  '.'
)[0];
```

My test can go through.  Could we please make this more permissive? No reason to let this break our test environment, this is regression.

PR #297, opened 2020-01-26
feat: add `animationTypeForReplace` option

satya164:
Currently, when a screen is replaced the new screen comes into focus with a push animation. However, sometimes you might want to customize how the animation looks like.

For example, when the user logs out, animating out the previous screen like pop feels more natural than doing a push animation with the sign in screen. The PR adds a new `animationTypeForReplace` option to control this. Specifying `animationTypeForReplace: 'pop'` will pop the previous screen, otherwise the new screen will be pushed like before.

**Before**

**After**


packages/stack/src/views/Stack/StackView.tsx
```diff
@@ -125,23 +131,41 @@ class StackView extends React.Component<Props, State> {
...
+              replacingRouteKeys = [
+                ...replacingRouteKeys,
+                previousFocusedRoute.key,
+              ];
+
+              closingRouteKeys = closingRouteKeys.filter(
+                key => key !== previousFocusedRoute.key
+              );
+
+              // Keep the old route in state because it's visible under the new route, and removing it will feel abrupt
```
osdnk:
```suggestion
              // Keep the old route in the state because it's visible under the new route, and removing it will feel abrupt
```

osdnk (approved):

PR #301, opened 2020-01-27
feat: add sidebar drawer type

NoemiRozpara:
to fix: now drawer opens always


packages/drawer/src/views/Drawer.tsx
```diff
@@ -81,7 +81,7 @@ type Props = {
   onGestureRef?: (ref: PanGestureHandler | null) => void;
   gestureEnabled: boolean;
   drawerPosition: 'left' | 'right';
-  drawerType: 'front' | 'back' | 'slide';
+  drawerType: 'front' | 'back' | 'slide' | 'sidebar';
```
EvanBacon, line 82: Following [Material design guidelines for drawers](https://material.io/components/navigation-drawer/#standard-drawer), `dismissible` or `permanent` might make sense.

NoemiRozpara: The problem is the behavior we want is dismissible on the small screen and permanent on the big ones. In the next iteration we want to add a prop to the state `isInitiallyVisible`, so maybe the drawer will be dismissible also on the big screens, but that haven't been decided yet. @satya164 do you have any thoughts on that?

NoemiRozpara: @EvanBacon @satya164 any update on this? I believe this is last thing to solve here, feel free to check the code :) Tested on iPad and iPhone 8 for now, new drawer type is not dismissible only on the big screen, as expected. Will test android soon.

osdnk:

PR #310, opened 2020-01-28
feat: add error if multiple instances of useLinking are used

osdnk:


babel.config.js
```diff
@@ -16,5 +16,7 @@ module.exports = {
   plugins: [
     '@babel/plugin-proposal-class-properties',
     '@babel/plugin-proposal-optional-chaining',
+    '@babel/transform-flow-strip-types',
```
satya164, line 19: Do we need this flow plugin?

packages/native/src/useLinking.tsx
```diff
@@ -40,6 +40,21 @@ export default function useLinking(
   ref: React.RefObject<NavigationContainerRef>,
   { prefixes, config, getStateFromPath = getStateFromPathDefault }: Options
 ) {
+  React.useEffect(() => {
+    const listenerRef = ref?.current?.__isListenerSetRef;
```
satya164: We need to use a variable in file scope to do it. Such a ref will handle 2 `useLinking` used for the same container, but if there were 2 containers (could be in a brownfield app), then we wouldn't catch it.

packages/native/src/useLinking.tsx
```diff
@@ -40,6 +40,21 @@ export default function useLinking(
...
+    if (listenerRef) {
+      if (listenerRef.current === false) {
+        listenerRef.current = true;
+      } else {
+        throw new Error(`Multiple instances of 'useLinking' are used`);
```
satya164:
```suggestion
        throw new Error("Looks like you are using 'useLinking' in multiple components. This is likely an error since deep links need to be handled only once.");
```

satya164 (changes requested):

packages/native/src/__tests__/useLinking.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  const wrapper = <Sample />;
+
+  expect(() => render(wrapper)).toThrowError(
+    `Multiple instances of 'useLinking' are used`
+  );
+});
+
+it('not throws if single instance of useLinking are used', () => {
```
WoLewicki: Can't these 2 tests be only one, where double `useLinking` throws and single doesn't? The preparation looks the same.

babel.config.js
osdnk: yup

PR #317, opened 2020-01-29
refactor: drop unmountInactiveScreens in favor of unmountOnBlur option

satya164:
The `unmountInactiveScreens` prop lets user unmount all inactive screens for the whole navigator when they go out of focus. It'll be better to have the option to do that per screen, so I have added the `unmountOnBlur` option instead.

To get the previous behaviour, user can specify the option in `screenOptions`.


Jalson1982:
This is a great option really!!! Hope this will be merged as soon as possible!

osdnk (approved):
I'm happy with this approach

Jalson1982:
Btw guys, is it possible to make that we have an option on bottom bar as we have in version 4 if we have stack navigator inside a tab bar, on tab bar press to go to an initial router of stack? This would be awesome too without resetting state...

satya164:
@Jalson1982 that happens by default if the tab is already focused. if you want it to happen all the time, you can add a listener on `tabPress` inside your stack navigator component and call `navigation.dispatch(StackActions.popToTop())` (`import { StackActions } from '@react-navigation/routers'`)

Jalson1982:
@satya164 
[export function ContactsScreen({ navigation }) {
  React.useEffect(() => {
    const subscribe = navigation.addListener('tabPress', e => {
      navigation.dispatch(StackActions.popToTop());
    });
    return subscribe;
  }, [navigation]);
  return (
    <Stack.Navigator headerMode="none" initialRouteName="Contacts" stackAnimation="flip">
      <Stack.Screen name="Contacts" component={ContactsList} />
      <Stack.Screen name="UserProfile" component={UserProfile} />
      <Stack.Screen name="Interests" component={Interests} />
    </Stack.Navigator>
  );
}](url)
Got error: 
The action POP_TO_TOP with payload undefined was not handled by any navigator.

satya164:
Try:

```js
if (props.route.state?.index > 0) {
  navigation.dispatch(StackActions.popToTop());
}
```

Jalson1982:
Strange still do not work. If I put console.log(props.route.state) it is undefined all time...

Jalson1982:
Ok I still have but now different error: undefined is not an object evaluating action.target

Jalson1982:
I fixed it, with this code it seems:
React.useEffect(() => {
    const subscribe = navigation.addListener('tabPress', e => {
      if (props.route.state?.index > 0) {
          navigation.popToTop()
      } 
    });
    return subscribe;
  }, [navigation,route]);

PR #322, opened 2020-01-30
feat: add initialRouteName property to config

WoLewicki:
Added `initialRouteName` property to config. When present, the screen of its name is always added to the state object on the proper level of nesting. There is no support for `initialRouteName` inside another `intialRouteName`, so if `initialRouteName` points to a navigator, it should be resolved properly by navigator itself.

### DIFF
--- a/packages/core/src/__tests__/getStateFromPath.test.tsx
+++ b/packages/core/src/__tests__/getStateFromPath.test.tsx
@@ -144,7 +144,6 @@ it('converts path string to initial state with config with nested screens', () =
   const path = '/foe/bar/sweet/apple/baz/jane?count=10&answer=42&valid=true';
   const config = {
     Foo: {
-      path: 'foo',
       screens: {
         Foe: 'foe',
       },
@@ -209,7 +208,6 @@ it('converts path string to initial state with config with nested screens and un
   const path = '/foe/baz/jane?count=10&answer=42&valid=true';
   const config = {
     Foo: {
-      path: 'foo',
       screens: {
         Foe: 'foe',
       },
@@ -263,14 +261,12 @@ it('handles nested object with unused configs and with parse in it', () => {
   const path = '/bar/sweet/apple/foe/bis/jane?count=10&answer=42&valid=true';
   const config = {
     Foo: {
-      path: 'foo',
       screens: {
         Foe: 'foe',
       },
     },
     Bar: 'bar/:type/:fruit',
     Baz: {
-      path: 'baz',
       screens: {
         Bos: 'bos',
         Bis: {
@@ -343,11 +339,9 @@ it('handles parse in nested object for second route depth', () => {
   const path = '/baz';
   const config = {
     Foo: {
-      path: 'foo',
       screens: {
         Foe: 'foe',
         Bar: {
-          path: 'bar',
           screens: {
             Baz: 'baz',
           },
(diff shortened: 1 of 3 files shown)


packages/core/src/getStateFromPath.tsx
```diff
@@ -207,13 +261,28 @@ function createNormalizedConfigs(
     // if an object is specified as the value (e.g. Foo: { ... }),
     // it has `path` property and
     // it could have `screens` prop which has nested configs
-    configs.push(createConfigItem(routeNames, value.path, value.parse));
+    if (value.initialRouteName) {
+      if (!value.screens?.[value.initialRouteName]) {
+        throw Error(
+          `No such route in ${value}'s config: ${value.initialRouteName}`
```
satya164: I think this is not needed. Initial route name shouldn't need an explicit path mapping

packages/core/src/__tests__/getStateFromPath.test.tsx
```diff
@@ -144,7 +144,6 @@ it('converts path string to initial state with config with nested screens', () =
   const path = '/foe/bar/sweet/apple/baz/jane?count=10&answer=42&valid=true';
   const config = {
     Foo: {
-      path: 'foo',
```
satya164: Do we need to change existing tests? This should be an additional feature, right?

WoLewicki: Yeah, I just did it for clarity that it is a navigator and there should be no need to navigate to it explicitly, but it doesn't break anything.

packages/core/src/getStateFromPath.tsx
WoLewicki: You are right, I will change it.

packages/core/src/getStateFromPath.tsx
```diff
@@ -110,47 +119,43 @@ export default function getStateFromPath(
...
-            },
-          };
-          helper = helper.routes[0].state as InitialState;
+    let state: InitialState = { routes: [] };
+    let routeName = routeNames.shift() as string;
+    let initialRoute = findInitialRoute(routeName, initialRoutes);
+
+    parseRoute(state, initialRoute, routeName, routeNames.length === 0, params);
+
+    if (routeName.length > 0) {
```
osdnk: ??
```suggestion
    if (routeNames.length > 0) {
```

packages/core/src/getStateFromPath.tsx
```diff
@@ -110,47 +119,43 @@ export default function getStateFromPath(
...
+        initialRoute = findInitialRoute(routeName, initialRoutes);
+        parseRoute(
+          helper,
+          initialRoute,
+          routeName,
+          routeNames.length === 0,
+          params
+        );
+        if (routeNames.length > 0) {
+          helper = helper.index
```
osdnk:
```suggestion
          helper = helper.routes[helper.index || 0].state as InitialState
```

osdnk: and pls use better name

packages/core/src/getStateFromPath.tsx
```diff
@@ -254,3 +272,56 @@ function findParseConfigForRoute(
...
+}
+
+function parseRoute(
+  state: InitialState,
+  initialRoute: string | undefined,
+  routeName: string,
+  isEmpty: boolean,
+  params?: Record<string, any> | undefined
+) {
+  if (isEmpty) {
```
osdnk, line 300: looks good, but again, make it please more descriptive. Also, could you try to omit to make side effects and use `return`?

WoLewicki: can do

satya164 (approved):

PR #326, opened 2020-01-30
feat: handle closing drawer before going back in useBackButton

osdnk:


packages/native/src/useBackButton.tsx
```diff
@@ -16,7 +16,9 @@ export default function useBackButton(
         }
 
         if (navigation.canGoBack()) {
-          navigation.goBack();
+          navigation.dispatch({ type: 'CLOSE_DRAWER' }, () =>
+            setImmediate(navigation.goBack)
```
satya164, line 21: I don't think it's desirable to both close the drawer and go back on back button press.

If there's such a use case, we can add an option in drawer navigator and internally it can listen to back button press and dismiss the drawer without. But I'm not convinced about this.

packages/core/src/types.tsx
```diff
@@ -322,9 +322,11 @@ type NavigationHelpersCommon<
    * The update function will receive the current state,
    *
    * @param action Action object or update function.
+   * @param onNotHandled callback after finished not handled transaction.
    */
   dispatch(
```
satya164, line 327: If rather not increase the API surface of dispatch by adding a callback here.

satya164 (changes requested):
Can you describe the use case for the PR?

Jalson1982:
@satya164 @osdnk  this should be good to solve in a way on android devices if a drawer is open and you press the android hardware back button actually to close drawer not to go back in stack navigation routes and in this case back button should not have any effect on stack navigation just simple close drawer menu.

osdnk:
Fixed on master

PR #338, opened 2020-02-02
fix: wrap navigators in gesture handler root

satya164:


example/package.json
```diff
@@ -21,13 +21,13 @@
     "react": "~16.9.0",
     "react-dom": "~16.9.0",
     "react-native": "~0.61.5",
-    "react-native-gesture-handler": "~1.5.3",
+    "react-native-gesture-handler": "~1.5.5",
```
osdnk, line 24: can you please consistently use "^"? There's no reason for ~
```suggestion
    "react-native-gesture-handler": "^1.5.5",
```

PR #344, opened 2020-02-03
refactor: rename NavigationNativeContainer to NavigationContainer

osdnk:


packages/native/src/index.tsx
```diff
@@ -1,6 +1,6 @@
 export * from '@react-navigation/core';
 
-export { default as NavigationNativeContainer } from './NavigationNativeContainer';
+export { default as NavigationContainer } from './NavigationNativeContainer';
```
satya164: Change the file name and component's name too (for devtools)

PR #349, opened 2020-02-04
feat: support ignoring empty path strings

WoLewicki:
In getPathFromState, If a path property for a route in the config object is an empty string, the part of the state object connected to that route is not concerned in the creation of URL string.


satya164 (approved):

packages/core/src/getStateFromPath.tsx
```diff
@@ -211,12 +211,14 @@ function createNormalizedConfigs(
...
-    configs.push(createConfigItem(routeNames, value));
+    if (value !== '') {
+      configs.push(createConfigItem(routeNames, value));
+    }
   } else if (typeof value === 'object') {
     // if an object is specified as the value (e.g. Foo: { ... }),
     // it can have `path` property and
     // it could have `screens` prop which has nested configs
-    if (value.path) {
+    if (value.path && value.path !== '') {
```
osdnk, line 221:
```suggestion
    if (value.path) {
```

PR #360, opened 2020-02-05
feat: useScrollToTop accepts an optional offset param  for scrollables where the "top" is not at 0 

christianjuth:
I have a FlatList component in my app that extends the full height of the device to achieve the iOS "frosted glass" effect. I had to do a bunch of tricks to get all the content to align correctly. One of those tricks was passing `contentOffset` to the FlatList to have it initialize to where I want the top of my list to be. Currently, there is no way to pass an offset value to `useScrollToTop` hook in React Navigation. I added a second parameter (after ref) that accepts an offset.

christianjuth/react-navigation.git#react-navigation-native-v5.0.0-gitpkg

```tsx
let contentInsets = {
  top: TopTabBar.useHeight({ safe: true }),
  bottom: Footer.useHeight({ safe: true })
};

let scrollIndicatorInsets = {
  top: TopTabBar.useHeight({ safe: false }),
  bottom: Footer.useHeight({ safe: false })
};

// tap tab to scroll to top
let ref = useRef();
useScrollToTop(ref, -contentInsets.top);

return(
  <FlatList
    contentContainerStyle={{
      paddingTop: 6,
      paddingBottom: 6
    }}
    contentInset={contentInsets}
    contentOffset={{
      y: -contentInsets.top,
      x: 0
    }}
    automaticallyAdjustContentInsets={false}
    scrollIndicatorInsets={scrollIndicatorInsets}
    contentInsetAdjustmentBehavior="never"
  />
);
```

Feb-05-2020 15-39-24


packages/native/src/useScrollToTop.tsx
```diff
@@ -82,20 +83,23 @@ export default function useScrollToTop(
           const scrollable = getScrollableNode(ref);
 
           if (isFocused && isFirst && scrollable && !e.defaultPrevented) {
-            if ('scrollToTop' in scrollable) {
-              scrollable.scrollToTop();
-            } else if ('scrollTo' in scrollable) {
-              scrollable.scrollTo({ y: 0, animated: true });
+            if ('scrollTo' in scrollable) {
```
christianjuth: I changed the order of these function calls so prioritizes scroll functions that support offset

packages/native/src/useScrollToTop.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { useNavigation, useRoute, EventArg } from '@react-navigation/core';
 
-type ScrollOptions = { y?: number; animated?: boolean };
```
christianjuth: Seems like this may have been a mistake and `y` should be renamed to `offset`

osdnk:

